PR #1, opened 2015-04-16
Moving Config classes into pihcore

mseaton:


mseaton:
@mogoodrich and @djazayeri , here is a pass at moving Config into pihcore.  Thoughts welcome.

api/src/main/java/org/openmrs/module/pihcore/config/ConfigDescriptor.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.List;
+
+/**
+ * Object that encapsulates the options that can be configured on a per-installation basis
+ */
+public class ConfigDescriptor {
+
+    public enum Site {
+        MIREBALAIS, LACOLLINE
```
djazayeri: I'm not sure how scalable it is to explicitly list out all our installations here, but it's not a blocker at this point.

PR #3, opened 2017-01-09
UHM-2709: Give residents permissions to print paper record labels

koolstens:


mogoodrich:
@koolstens looks good, just one comment: can we change role name from "LEGACY_PRINT_LABELS" to just "PRINT_LABELS".   "Legacy" was just a term I was adding to the "old" privileges that we are hoping to phase out eventually.  

I think you can just make this one change, commit it, and then push it up to your repo and it will automatically add it to the pull request.  So once you make the change you'd first add the files to the list of files to commit with the code:

git add <filenames_to_add>

then you would commit the code to your local repository:

git commit -m "#2709: Give residents permissions to print paper record labels"

and then finally push it up to your repo:

git push koolstens master  (<-- I think that is right, I always forget the order)

PR #7, opened 2018-07-20
MEX-27: Remove check-in feature

brandones:
https://tickets.pih-emr.org/browse/MEX-27


brandones:
NB: This causes a transient NullPointerException when going to the patient dashboard of a patient who has been checked in before, caused by a cache failure. It resolves on hitting "back" and trying again, and doesn't come back (as far as I can tell).

PR #11, opened 2018-08-17
MEX-6: Address should have fields "community" and "directions"

brandones:
https://tickets.pih-emr.org/browse/MEX-6

This adds address hierarchy entries; changes address format.

This includes "Otro" (Other) as an option for Community.

This also seems to add a field called something like "Rapidly search for an address" that seems quite redundant with the Community field itself. Eventually we'll probably want to get rid of that.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/mexico/MexicoAddressBundle.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/mexico/MexicoAddressBundle.java
@@ -1,0 +1,39 @@
+package org.openmrs.module.pihcore.deploy.bundle.mexico;
+
+import org.openmrs.module.addresshierarchy.AddressField;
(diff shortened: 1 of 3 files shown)


mogoodrich:
@brandones didn't realize your hierarchy was so simple!

Here's where the "shortcut" search gets set up (which renders the "Rapidly search for an address") :
https://github.com/PIH/openmrs-module-mirebalais/blob/master/api/src/main/java/org/openmrs/module/mirebalais/apploader/apps/PatientRegistrationApp.java#L722

Long-term, we could configure this by country, but for now if you wrap this line in a "!config.getCountry().equals(ConfigDescriptor.Country.MEXICO)" I believe that will disable the shortcut in your context.

brandones:
Fantastic, thanks Mark. Unfortunately, I just found out about some legal requirements that make our address hierarchy a lot less simple. So I'll be creating another PR to add a few new fields soon.

PR #16, opened 2018-09-13
MEX-99 Remove History and Exam from templates

brandones:
https://tickets.pih-emr.org/browse/MEX-99


brandones:
(Exam had been removed by a previous change)

PR #19, opened 2018-09-30
MEX 7: Diabetes form

brandones:
https://tickets.pih-emr.org/browse/MEX-7

Also includes some stuff from the default NCD form, some of which we'll want to keep.


dsurrao:
@brandones before I merge, i wanted to ask about the use of PIH codes, e.g., PIH:10555, is there any way to avoid using IDs? @lnball ?

brandones:
The PIH:XXXXX codes are from code I copy-pasted -- I'd be happy to add actual name mappings if that's the practice we're moving toward.

brandones:
Ping?

lnball:
Feel free to use any PIH or CIEL SAME-AS mappings.  My recommended and current practice:


  *   First choice - use CIEL (SAME-AS) mappings if available since that is best for the larger OpenMRS community.  Others can reuse the htmlform if they are using the CIEL dictionary.
  *   Second choice - If the concept doesn't have a corresponding CIEL concept/mapping, I use the PIH:name (ie. 'PIH:Anemia') mapping since it makes for readable htmlforms.
  *   Third choice - PIH:number mappings (ie. 'PIH:12').  Not readable, but will work.


Ellen


________________________________

mseaton:
@lnball and @dsurrao - I think @brandones is waiting on us to merge this pull request it, or indicate if there is a reason not to.  Ellen, are you asking Brandon to replace his number mappings with ciel or name mappings? Or is this good to go?

lnball:
What's the process to merge and close?  Who?  And so we're not stepping on each others toes.

PR #20, opened 2018-10-09
MEX-149: Add boolean person attributes to Registration: immigrant, indigenous, active casefinding

brandones:
https://tickets.pih-emr.org/browse/MEX-149

I tried to create a new registration form section for "active casefinding" since it's not really demographic information, but I couldn't get the new section to appear. Not sure what I was missing. I've left it in Demographics for now.


brandones:
This has been updated to use concepts instead of patient attributes.

PR #36, opened 2019-04-09
MEX-207 Enable Mental Health program for CES

brandones:
MEX-207 Enable Mental Health program for CES
https://tickets.pih-emr.org/browse/MEX-207

Also includes some comments -- the beginnings of documenting what our components actually do.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
@@ -21,10 +21,11 @@ public class Components {
         public static final String VCT = "vct";
         public static final String MCH = "mch";  // program
         public static final String ZIKA = "zika";  // program
-        public static final String MENTAL_HEALTH = "mentalHealth";
-        public static final String MENTAL_HEALTH_PROGRAM = "mentalHealthProgram";
-        public static final String HYPERTENSION_PROGRAM = "hypertensionProgram";
-        public static final String DIABETES_PROGRAM = "diabetesProgram";
+        public static final String MENTAL_HEALTH= "mentalHealth";  // mentalHealthForm + mentalHealthProgram
+        public static final String MENTAL_HEALTH_FORM = "mentalHealthForm";  // used by Haiti Mental Health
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -21,10 +21,11 @@
...
         public static final String ZIKA = "zika";  // program
-        public static final String MENTAL_HEALTH = "mentalHealth";
-        public static final String MENTAL_HEALTH_PROGRAM = "mentalHealthProgram";
-        public static final String HYPERTENSION_PROGRAM = "hypertensionProgram";
-        public static final String DIABETES_PROGRAM = "diabetesProgram";
+        public static final String MENTAL_HEALTH= "mentalHealth";  // mentalHealthForm + mentalHealthProgram
+        public static final String MENTAL_HEALTH_FORM = "mentalHealthForm";  // used by Haiti Mental Health
+        public static final String MENTAL_HEALTH_PROGRAM = "mentalHealthProgram";  // used by Haiti and Mexico
+        public static final String HYPERTENSION_PROGRAM = "hypertensionProgram";  // Mexico program
+        public static final String DIABETES_PROGRAM = "diabetesProgram";  // Mexico program
```
mogoodrich: Thanks for added thing much-needed comments... I'd tweak the comments a bit though:

// mental health form (currently used in Haiti)
// mental health program (requires MentalHealthProgramBundle to be installed) (currently used in Haiti and Mexico)
// hypertension program (requires HypertensionProgramBundle to be installed) (currently used in Mexico)

etc...  ideally each of these is country-independent... I know that  it's possible we will have to make some country-dependent (maybe MCH?) but would like to make sure people think of them as country-independent as long as possible...

There could be an argument for not including the "currently used in" clause because if we don't update they will grow stale, but I'm find with that....

brandones: Nice, thanks for the clarification. I’m going to advocate eliding “currently” solely for brevity, otherwise with you 100%.

PR #37, opened 2019-04-11
Enable edTriage, edTriageQueue and edConsult for Kouka

alaboso:
Enable edTriage, edTriageQueue and edConsult for Kouka -- Harper -- JJ Dossen

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
@@ -109,6 +109,8 @@ public class LocationTagSetup {
 
         Field locationField = LiberiaLocations.class.getField(config.getSite().name().toString());
         LocationDescriptor location = (LocationDescriptor) locationField.get(LocationDescriptor.class);
+        // Enable edTriage, edTriageQueue and edConsult for Kouka -- Harper -- JJ Dossen
+        List<LocationDescriptor> harper = Arrays.asList(LiberiaLocations.HARPER);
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
```diff
@@ -109,6 +109,8 @@ private static void setupLocationTagsForLiberia(LocationService locationService,
 
         Field locationField = LiberiaLocations.class.getField(config.getSite().name().toString());
         LocationDescriptor location = (LocationDescriptor) locationField.get(LocationDescriptor.class);
```
mogoodrich: So I believe what this does is fetch the name of the current server site (either Pleebo or Harper) and then pulls that location descriptor from the LiberiaLocations class.  So on the Pleebo server, all these tags get added to Pleebo, and on the Harper server, all these tags get added to Harper.

With what you've added, the ED location tags will get added to the "harper" location on both the Pleebo and Harper servers, I believe... I don't think that's a big deal (I think the main concern is we don't want Harper to be tagged as a "Login location" on Pleebo or vice versa, or both will appear on the Login screen as options)  but if we can maintain the existing pattern it's probably better for consistency? Or we could potentially change it up so that we enable all the tags on for both locations on both servers, except for login?

 (If the ED component isn't enabled on Pleebo it still won't show up, even if the tags are there, so we don't need to worry about ED mistakenly being enabled on Pleebo.)

alaboso: Thanks for the correction. I have made the correction

PR #40, opened 2019-04-12
MEX-161, MEX-165, MEX-202: no imperial units; subjective box on consult form

brandones:
https://tickets.pih-emr.org/browse/MEX-165
https://tickets.pih-emr.org/browse/MEX-202
https://tickets.pih-emr.org/browse/MEX-161

### DIFF
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -54,6 +54,7 @@ pihcore.visitNote.visitTemplate=Visit Template
 pihcore.visitNote.clinicalNotes=Clinical Notes
 pihcore.visitNote.duringCurrentConsult=During current consult
 
+pihcore.chiefComplaint.title=Chief Complaint
 pihcore.subjective.title=Subjective
 
 pihcore.feeding.history.title=Feeding History
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -54,7 +54,7 @@ pihcore.visitNote.visitTemplate=Visit Template
 pihcore.visitNote.clinicalNotes=Clinical Notes
 pihcore.visitNote.duringCurrentConsult=During current consult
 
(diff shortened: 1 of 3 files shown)


api/src/main/resources/messages.properties
```diff
@@ -54,7 +54,7 @@ pihcore.visitNote.visitTemplate=Visit Template
 pihcore.visitNote.clinicalNotes=Clinical Notes
 pihcore.visitNote.duringCurrentConsult=During current consult
 
-pihcore.chiefComplaint.title=Chief Complaint
```
mogoodrich: Why remove this? I believe we use it elsewhere....

omod/src/main/webapp/resources/htmlforms/mexico/vitals.xml
```diff
@@ -519,20 +268,6 @@
...
-                        <input type="text" id="temperature_f" size="5" min="77.0" max="109" class="number numeric-range fahrenheit"/>
-                        <span class="append-to-value"><uimessage code="emr.units.degreesFahrenheit"/></span>
-                        <span class="error field-error" style="display: none"></span>
-                    </span>
-                </p>
-                <p class="left">
-                    <uimessage code="mirebalais.vitals.or"/>
-                </p>
-            </ifMode>
-
```
mogoodrich: I didn't review this too deeply, but it looks good, I'll trust you pulled the right code... (or at least will test to make sure it works).

Nice to get rid of this, if at least in Mexico.  Story behind this... when we opened Mirebalais we all agreed to standardize on metric, and then when Renee (former MedInfo member) got to Mirebalais to do training she discovered that of the donated scales, some measured in metric and other in imperial...

api/src/main/resources/messages.properties
brandones: Oh shoot, good catch!

PR #41, opened 2019-04-12
MEX-208 Create/Enable Maternal Health program for CES

brandones:
MEX-208 Create/Enable Maternal Health program for CES
https://tickets.pih-emr.org/browse/MEX-208


api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/haiti/HaitiMetadataToInstallAfterConceptsBundle.java
```diff
@@ -2,15 +2,27 @@
...
 // TODO: incorporate all PIH metadata installation into this module, remove Mirebalais Metadata module?
 @Component
-@Requires( { PihProgramsBundle.class, OrderEntryConcepts.class } )
+@Requires( { OrderEntryConcepts.class,
+        HIVProgramBundle.class,
+        MentalHealthProgramBundle.class,
+        MCHProgramBundleZL.class,
+        NCDProgramBundle.class,
+        OncologyProgramBundle.class,
+        ZikaProgramBundle.class } )
```
mogoodrich: Ideally, this could still depend on the PihProgramsBundle, and the MCHProgramBundleZL would just be installed in on top of the MCHProgram bundle, but maybe there's something funky here since it actually modifying the program component (by adding workflows).  Did you try this?  (If not, no worries, I could potentially test myself rather than you having to set up a whole Haiti dev environment, but wanted to check first if you had tried this)

brandones: Well, I've accidentally tried to have it load two programs with the same UUID (by creating a new program but forgetting to change the UUID from the one I copied), and it crashes, complaining that there's already a program with that UUID. I wasn't able to figure out a way to specify an "override" to @Requires. The "subclassing" idea also didn't pan out as cleanly as I would have liked, since the MCH ProgramDescription object also required some ZL-specific changes.

I suspect a good general solution would be something along the lines of a ProgramFactory. But that's probably outside the scope of this ticket, more in the territory of "refactor." But I dunno, what do you think? Worth it? I'd call it a day of work.

mogoodrich: Not worth it at this point, though still work thinking about the overall refactor...
I would think that it *might* work if the program had the same name...

That being said, I'll merge this in, and then might play around with it later today if I have time...

PR #48, opened 2019-04-30
LIB:  Added JJ Dossen as site/facility supported and updated patient …

alaboso:
- Added JJ Dossen as site/facility supported 
- Updated patient identifier and identifier types & sources

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/metadata/liberia/LiberiaLocations.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/metadata/liberia/LiberiaLocations.java
@@ -17,6 +17,13 @@ public class LiberiaLocations {
 
     public static LocationDescriptor HARPER = new LocationDescriptor() {
         public String uuid() { return "6fa7bab5-1ca5-4b77-8c31-80c9589af952"; }
+        public String name() { return "JJ Dossen (Kouka)"; }
+        public String description() { return "JJ Dossen (Kouka)"; }
+        public List<LocationTagDescriptor> tags() { return Collections.emptyList(); }
+    };
+
+    public static LocationDescriptor JJ_DOSSEN = new LocationDescriptor() {
+        public String uuid() { return "06347c45-6a8e-11e9-824c-645d86728797"; }
         public String name() { return "JJ Dossen"; }
         public String description() { return "JJ Dossen"; }
         public List<LocationTagDescriptor> tags() { return Collections.emptyList(); }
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/module/pihcore/metadata/liberia/LiberiaLocations.java
```diff
@@ -17,6 +17,13 @@
 
     public static LocationDescriptor HARPER = new LocationDescriptor() {
         public String uuid() { return "6fa7bab5-1ca5-4b77-8c31-80c9589af952"; }
+        public String name() { return "JJ Dossen (Kouka)"; }
+        public String description() { return "JJ Dossen (Kouka)"; }
```
mogoodrich: We don't need a separate location for Kouka, the staging server... I would say just go ahead and change the "HARPER" variable and the "HARPER" site to "JJ_DOSSEN"... we were intending "Harper" to refer to "JJ Dossen" but we should switch to the clearer name.  It's perfectly fine for Kouka to use the "real" JJ Dossen location as well and basically be a staging server for JJ Dossen.

alaboso: I'd rather leave it as is because there are instances where we want to test a feature/module before actual activation. e.g. ED Triage/Queue. This we do in Kouka without necessarily availing the functionality in prod JJ_DOSSEN.

mogoodrich: Absolutely, we need a way to turn on functionality on Kouka but not on JJ Dossen, but we should be able to do this via the components enabled in pih-config-liberia-kouka vs pih-config-liberia-jjdossen... both servers can share the same location as long as their component configurations are different... this is what we do for our Haiti servers... for instance, our CI staging server is configured with the "Thomonde" location but has extra features turned on...

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/openmrs/files/config/pih-config-haiti-thomonde.json

Both the Thomonde production server and ci.pih-emr.org use this config:

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/openmrs/files/config/pih-config-haiti-thomonde.json

But then we override a custom set of components for CI:

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/openmrs/files/config/pih-config-haiti-thomonde-ci.json

https://github.com/PIH/mirebalais-puppet/blob/master/hieradata/thomonde.pih-emr.org.yaml#L4
https://github.com/PIH/mirebalais-puppet/blob/master/hieradata/ci.pih-emr.org.yaml#L3

alaboso:
- I have updated the files, also updated puppet configs. 
- I think the simplest way is to have a new set of configs for JJ_Dossen extending from _liberia-harper_ so that we have _liberia-harper-kouka_ and _liberia-harper-jjdossen_ as both use same configs except (de)activated components. 
- What I noted though, is that the location tags will be active for ED Triage/Queue in both locations but since ED component isn't activated in _liberia-harper-jjdossen_, this location will not have this module active, and it's what's intended.

PR #51, opened 2019-05-10
MEX-323 Asthma nighttime symptom labels are wrong

brandones:
https://tickets.pih-emr.org/browse/MEX-323


brandones:
Ellen solved in a better way.

PR #53, opened 2019-05-13
MEX-148 Consult form should display forms for programs depending on program enrollment status

brandones:
https://tickets.pih-emr.org/browse/MEX-148

Depends on https://github.com/openmrs/openmrs-module-htmlformentry/pull/147

Also move Asthma above Diabetes


brandones:
How do I tell when the build has gone through?

brandones:
For posterity: The build happens on the PIH Bamboo server.

PR #62, opened 2019-06-03
MEX-301 Glucose question should be asked with "Fasting for glucose test"

brandones:
https://pihemr.atlassian.net/browse/MEX-301


brandones:
Thanks Cosmin!

brandones:
@lnball FYI. As noted in the comments on https://pihemr.atlassian.net/browse/MEX-301, I wasn't able to model as we had discussed. We can work on this more later.

PR #68, opened 2019-06-11
MEX-121 Should be able to add CHW as relationship from dashboard

brandones:
MEX-121 Should be able to add CHW as relationship from dashboard
https://pihemr.atlassian.net/browse/MEX-121


mogoodrich:
LGTM

PR #71, opened 2019-06-18
MEX-362 Height/weight graph on main patient dashboard

brandones:
https://pihemr.atlassian.net/browse/MEX-362


mogoodrich:
@brandones  looks like this likely can be closed if we go with the other approach?

brandones:
Per comment on https://github.com/PIH/openmrs-module-mirebalais/pull/54#pullrequestreview-255311104, I'll merge and then try to make sure we fix this sooner rather than later.

PR #74, opened 2019-06-26
Make a bunch of improvements to the Mexico consult form

brandones:
In brief:
- chill out the error color a bit
- make it so that valid values don't turn green (MEX-322 Research disabling red/green shading on numeric form fields)
- add last subjective and lab order text to the left of subjective box (MEX-253 Previous "Subjective" field entry should be visible next to form entry box, MEX-273 Contents of previous "Lab tests" box should appear next to "Subjective")
- add previous values to asthma (except for "nighttime symptoms," because the diagnosis constructs make this impossible. I also wonder about how one can represent these in dashboard widgets.)
- add previous values to diabetes
- add a button to epilepsy that the user needs to click in order to change the baseline value
- and, the best part: incremental show/hide buttons for medications. I'm opening another PR for the same change applied to `section-plan.xml` for HUM et al.

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
+++ b/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
@@ -34,9 +34,13 @@
         }
 
         .field-error {
-            color: #ff6666;
+            color: #ffc0b5;
             font-size: 1.1em;
             display: block;
+        }
+
+        .legalValue {
+            background-color: white !important;
         }
 
         .two-columns {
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -34,11 +34,15 @@
         }
 
         .field-error {
-            color: #ff6666;
+            color: #ffc0b5;
```
mogoodrich: Is this the fix for the issue making abnormal values green?  If this gets the job done, that's fine, but I've got no problem with making a change to how the highlighting is done in HFE directly... it seems like what is currently happening is wrong, and I don't think there was any particular thought/strategy behind it, likely just something whoever was developing the initial ticket thought was nice to add.

brandones: This one just makes the red a little less aggressive.

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -34,11 +34,15 @@
...
             font-size: 1.1em;
             display: block;
         }
 
+        .legalValue {
+            background-color: white !important;
```
brandones: this is what stops the fields from turning green when valid

PR #75, opened 2019-06-26
Make CES locations PROVIDER_MANAGEMENT locations so they can use chwApp

brandones:


api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
```diff
@@ -74,23 +74,18 @@ private static void setupLocationTagsForSierraLeone(LocationService locationServ
...
 
-        Field locationField = MexicoLocations.class.getField(config.getSite().name().toString());
+        // Get the current location -- each site only has one available
+        Field locationField = MexicoLocations.class.getField(config.getSite().name());
         LocationDescriptor location = (LocationDescriptor) locationField.get(LocationDescriptor.class);
 
-        // Tags for our hub in Jaltenango
-        List<LocationDescriptor> jalte = Arrays.asList(MexicoLocations.JALTENANGO);
-        setLocationTagsFor(locationService, LocationTags.LOGIN_LOCATION, jalte);
-        setLocationTagsFor(locationService, LocationTags.MEDICAL_RECORD_LOCATION, jalte);
```
mogoodrich: I assume that removing these and the NCD consult location was intentional?

brandones: Yes—there is no Jaltenango location. Actually, I should remove that from MexicoLocations as well.

PR #77, opened 2019-06-26
Make medication list increment

brandones:
Applying the medication form improvement from https://github.com/PIH/openmrs-module-pihcore/pull/74 to PrimaryCare and NCD

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/section-plan.xml
+++ b/omod/src/main/webapp/resources/htmlforms/section-plan.xml
@@ -64,6 +64,9 @@
     <ifMode mode="VIEW" include="false">
         <script type="text/javascript">
 
+            // I don't really understand why this assignment `$j = jq` is required, but for some
+            // reason a bunch of Javascript gets included in the page that expects `$j` to exist,
+            // but doesn't define `$j` itself. *shrug emoji*
             var $j = jQuery;
 
             htmlForm.getBeforeValidation().push(function() {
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/resources/htmlforms/section-plan.xml
```diff
@@ -123,45 +123,54 @@
...
-                var hideOtherMeds = function() {
-                    jq('#medication-2').hide();
-                    jq('#medication-3').hide();
-                    jq('#medication-4').hide();
-                    jq('#medication-5').hide();
-                    jq('#medication-6').hide();
-                    jq('#medication-7').hide();
-                    jq('#medication-8').hide();
+                var initMeds = function () {
+                    for (var i = 1; i &lt;= 8; i++) {
```
mogoodrich: Wait, so this is much nicer because it doesn't require hard-coding all the hides, etc, for each field, but does it still limit the number of medications to 8?  this was just an arbitrary restriction if we are now able to remove it.

We we also need to make sure we do some good testing once this is merged.

brandones: Oh yeah! Lemme bump that number to something more unlikely to ever be hit. Like 30? I'll factor it out into a variable, too.

brandones:
Ok, updated both sections-plan and mexico consult to have a MAX_MEDS parameter set to 30.

brandones:
Yeah, also because my connection is spotty (to put it mildly) -- if you could merge and test, that would be great. You did try it locally, right? I wasn't able to get a HUM server working right.

lnball:
I assume this improvement will appear for Haiti too?  Nice.  Let me know when I can see this on humci.

PR #78, opened 2019-06-27
SL-88 Create new encounter type for Outpatient Visit

brandones:
https://pihemr.atlassian.net/browse/SL-88

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/EncounterTypeBundle.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/EncounterTypeBundle.java
@@ -49,6 +49,8 @@ public class EncounterTypeBundle extends AbstractMetadataBundle {
         install(EncounterTypes.VACCINATION);
         install(EncounterTypes.ECHOCARDIOGRAM);
         install(EncounterTypes.MEXICO_CONSULT);
+        install(EncounterTypes.SIERRA_LEONE_OUTPATIENT_INITIAL);
+        install(EncounterTypes.SIERRA_LEONE_OUTPATIENT_FOLLOWUP);
 
         uninstall(possible(EncounterType.class, AttachmentsConstants.ENCOUNTER_TYPE_UUID), "not used");
         uninstall(possible(EncounterType.class, EncounterTypes.MEXICO_CLINIC_VISIT.uuid()), "replaced with Mexico Consult");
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/EncounterTypeBundle.java
```diff
@@ -49,6 +49,8 @@ public void install() throws Exception {
         install(EncounterTypes.VACCINATION);
         install(EncounterTypes.ECHOCARDIOGRAM);
         install(EncounterTypes.MEXICO_CONSULT);
+        install(EncounterTypes.SIERRA_LEONE_OUTPATIENT_INITIAL);
+        install(EncounterTypes.SIERRA_LEONE_OUTPATIENT_FOLLOWUP);
```
mogoodrich: I think it makes sense to start breaking these out into implementation-specific bundles... do you have a argument as to why not do to this?

Seems weird to have Sierra Leone encounter types in Haiti or Mexico, etc...

omod/src/main/webapp/resources/scripts/visit/encounterTypeConfig.js
```diff
@@ -5,11 +5,6 @@ angular.module("encounterTypeConfig", ["constants"])
         var hfeSimpleEditUrl = "/htmlformentryui/htmlform/editHtmlFormWithSimpleUi.page?patientId={{encounter.patient.uuid}}&encounterId={{encounter.uuid}}&returnUrl={{returnUrl}}&breadcrumbOverride={{breadcrumbOverride}}";
         var hfeStandardEditUrl = "/htmlformentryui/htmlform/editHtmlFormWithStandardUi.page?patientId={{encounter.patient.uuid}}&encounterId={{encounter.uuid}}&returnUrl={{returnUrl}}&breadcrumbOverride={{breadcrumbOverride}}";
 
-        // if use has only "enterConsultNote", must be an active visit; retroConsultNote or retroConsultNoteThisProviderOnly required for retro visits
-        // TODO do we use this anywhere?
-        var standardConsultNoteRequire = "(user.hasPrivilege('Task: emr.enterConsultNote') && visit && !visit.stopDatetime) || (user.hasPrivilege('Task: emr.retroConsultNote') || user.hasPrivilege('Task: emr.retroConsultNoteThisProviderOnly'))"
```
mogoodrich: Is there reason you removed this?  (Was it dangled code no longer being used?)

brandones: ` // TODO do we use this anywhere?`... the answer appears to be no :)

api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/EncounterTypeBundle.java
brandones: Yes, I think that is definitely the right path. Call it a near-future refactor?

brandones:
Merging in order to un-block Martin.

PR #79, opened 2019-06-28
UHM-3863 Liquibase code to migrate the NCD diagnoses to use the const…

lnball:
@brandones Would you be so kind to review my migration code?  I'd be glad to explain further.

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -204,5 +204,257 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20190625-a-migrate-diabetes2-to-diabetes2InsulinDependent" author="ball">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                select count(*) from concept_name
+                 where name like 'Type 2 diabetes requiring insulin'
+                   and locale = 'en'
+                   and locale_preferred = 1
+                   and voided = 0;
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            UHM-3923 Migrate Diabetes 2 to Diabetes Type 2 requiring insulin for NCD encounters
+            For Version 2, Diabetes Type 2 requiring insulin and not insulin-dependent were introduced.
+            Darius Fenelon wanted to migrate all the general Type 2 Diabetes to requiring insulin data
+            since that's more common
+
+            Only run this when this concept is available (ie. Haiti)
+        </comment>
+        <sql>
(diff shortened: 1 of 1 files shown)


brandones:
Sure! Will this code just be run once, or will it live forever as additional liquibase steps?

lnball:
It will run once @brandones

mogoodrich:
Yes, @brandones  if you look in the OpenMRS DB you see there's a liquibase table that tracks which changesets have already been run to assure that the same changeset doesn't run more than once.

lnball:
@brandones It is a doozy.  And a good exercise to prove the rationale for good concept modeling from the start.  I wrote V1 and learned much before doing V2.  When in doubt -- use an obsgroup for valuable data.

As for hitting this sql if someone adds that concept, I have thought of that.  Not sure how to avoid -- unless I can have 2 preconditions -- one for Haiti (where the concept exists) and one for the concept.  I don't know how to do that.

lnball:
Do **not** merge.  I need to change the code to migrate Type 2 DM to Type 2 DM on oral drugs.

brandones:
Yep, per our discussion in the office last month, it should be left to the person who wrote the code to merge, since they'll have the best idea of when the right moment is, and the order in which things need to be done. Let me know when it's ready for another look.

brandones:
> @brandones It is a doozy. And a good exercise to prove the rationale for good concept modeling from the start. I wrote V1 and learned much before doing V2. When in doubt -- use an obsgroup for valuable data.

I'd be really interested in hearing the story of what prompted this transition for Haiti. My experience with obsgroups has basically been that they make it impossible to use that data within OpenMRS, e.g. for graphs, ObsByEncounter, velocity lookups, etc.

> As for hitting this sql if someone adds that concept, I have thought of that. Not sure how to avoid -- unless I can have 2 preconditions -- one for Haiti (where the concept exists) and one for the concept. I don't know how to do that.

```
SELECT COUNT(*) FROM global_property gp
CROSS JOIN concept_name cn
WHERE gp.property_value LIKE '%haiti%'
    AND gp.property ='layout.address.format';
    AND cn.name LIKE 'Type 2 diabetes requiring insulin'
    AND cn.locale = 'en'
    AND cn.locale_preferred = 1
    AND cn.voided = 0;
```
?

As an aside, @mseaton , this seems to relate to the question of whether or not `pih.config` should be a global property. The SQL we're using to check the site here, with `layout.address.format`, is pretty fragile and awkward.

PR #85, opened 2019-07-16
UHM-3937 Modified the history section for NCD V2

lnball:
@brandones Do you have time for this?  Otherwise I'll ask someone else.

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/section-history.xml
+++ b/omod/src/main/webapp/resources/htmlforms/section-history.xml
@@ -169,14 +169,17 @@
         </h1>
     </ifMode>
 
-    <section id="chiefComplaint" sectionTag="fieldset" headerTag="legend" headerStyle="title"
-             headerCode="pihcore.chiefComplaint.title">
-        <div class="section-container">
-            <p>
-                <obs conceptId="CIEL:160531" style="textarea" id="chief-complaint"/>
-            </p>
-        </div>
-    </section>
+    <!-- Show chief complaint if encounter type is NOT NCD Intake -->
+    <includeIf velocityTest="$encounter.encounterType.uuid != 'ae06d311-1866-455b-8a64-126a9bd74171'" >
+        <section id="chiefComplaint" sectionTag="fieldset" headerTag="legend" headerStyle="title"
+                 headerCode="pihcore.chiefComplaint.title">
+            <div class="section-container">
+                <p>
+                    <obs conceptId="CIEL:160531" style="textarea" id="chief-complaint"/>
+                </p>
+            </div>
+        </section>
+    </includeIf>
 
     <section id="complaint" sectionTag="fieldset" headerTag="legend" headerStyle="title"
              headerCode="pihcore.history.presentingHistory.title">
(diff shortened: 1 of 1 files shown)


brandones:
Sure taking a look now

omod/src/main/webapp/resources/htmlforms/section-history.xml
```diff
@@ -192,25 +195,25 @@
         <section id="complaint" sectionTag="fieldset" headerTag="legend" headerStyle="title"
```
brandones: Hmmm, I know this isn't part of your addition, but "complaint" is a duplicate ID. We should fix at some point.

omod/src/main/webapp/resources/htmlforms/section-history.xml
```diff
@@ -192,25 +195,25 @@
...
-            <div class="two-columns">
                 <p>
                     <label>
-                        <uimessage code="pihcore.refer.referBy"/>
+                        <uimessage code="pihcore.refer.date"/>
                     </label>
+                    <obs conceptId="CIEL:163181" allowTime="false" />
+                </p>
+
+            <div class="two-columns">
```
brandones: We seem to have some weird indentation going on... would you mind doing an auto-reformat prior to merging?

omod/src/main/webapp/resources/htmlforms/section-history.xml
```diff
@@ -464,50 +467,51 @@
...
-
-                            <!-- DPA -->
-                            <p>
-                                <label>
-                                    <uimessage code="pihcore.pregnancy.dueDate"/>
-                                </label>
-                                <obs conceptId="CIEL:5596" allowFutureDates="true" />
-                            </p>
+                    <!-- Pregnant -->
+                    <p>
```
brandones: The <p> tag shouldn't ever contain block-level elements (including <div>). I see that this is just a reformat, so... we should fix this at some point :)

lnball: I'm not up on this -- <p> with <div> tags.  If it looks good on the page, isn't that enough?  Can you recommend an improvement or resource with nice formatting @brandones ?

brandones: (I assume the big spaces in your comment are because you tried to type HTML, which GitHub is choosing to render rather than display plaintext. If you want to show HTML code you need to wrap it in backticks. Then
```
`<div>`
```
will render as
`<div>`).

> If it looks good on the page, isn't that enough?

Fair question! Apologies in advance if some of this you already know.

So, there's a group called [W3C](https://www.w3.org/Consortium/) that sets standards about web technologies. They set standards about HTML code. They define the semantic rules of the language -- they say "this is what X means, this is what X means with Y." All the browsers refer to that ruleset when developing their HTML renderer. They also say "you're not allowed to put a Z in a Y," which implies "if you do, we make no guarantees whatsoever about what the browser will do with that." It yields "undefined behavior." 

Putting block-level elements like `<div>` inside of `<p>` is forbidden, and its behavior is undefined. Browsers can choose to do whatever they like with that content. They might pretend there is no `<p>`. They might treat it like a `<div>` tag (which is probably what you should be using in all of these instances). They might apply the browser default styling for `<p>`. They might not. They might do inexplicable, bizarre things with the content. Different versions of the same browser may do radically different things with the content. The behavior is undefined.

This is the main reason we should try to write standards-compliant code.

This connects to a broader conversation about code quality -- the idea that whether something seems to work right is not a sufficient condition for it being good code, and that we should be striving and working together to produce good code.

PR #86, opened 2019-07-17
UHM-3953 - modifying liquibase changeset for faster performance .

mseaton:


lnball:
@mseaton Your change seems good; however, I have not used the validChecksum for a liquibase changeset.  How does that work?  I thought that I needed to replace the changeSet id (20190712-migrate-ncd-dm-diagnoses-to-construct) with a new one (20190718-migrate-ncd-dm-diagnoses-to-construct) or it would throw an error.  It should not be a problem to rerun on previously migrated code since it checks for the obsgroup and that should already exist.

@mogoodrich ?

PR #87, opened 2019-07-18
UHM-3879 Show gestational diabetes only for females; Also reformatted…

lnball:
… the code with intellij

One small section on a htmlform for code review.  It should be quick.  The only time-consuming part is that @brandones suggested I auto-format the code — so it's a small fix but many cosmetic changes to the file.

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/section-ncd.xml
+++ b/omod/src/main/webapp/resources/htmlforms/section-ncd.xml
@@ -219,7 +219,7 @@
 
             <!-- Program enrollment and set workflow state-->
             <p>
-                <enrollInProgram programId="NCD" />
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/htmlforms/section-ncd.xml
```diff
@@ -416,34 +415,60 @@
...
                     </th>
                     <th>
                         <label>
-                            <uimessage code="pihcore.suggestedTreatment.label" />
+                            <uimessage code="pihcore.suggestedTreatment.label"/>
                         </label>
                     </th>
                 </tr>
 
-                <tr>
```
brandones: Just for posterity: this block had a functional change

PR #88, opened 2019-07-20
Sl04pihcore

lebbiesahr:
Outpatient changes relating to ticket SL-112


omod/src/main/webapp/resources/htmlforms/sierra_leone/section-plan.xml
```diff
@@ -179,288 +179,63 @@
...
     </section>
 
-    <section id="radiology-orders" sectionTag="fieldset" headerTag="legend" headerStyle="title"
-             headerCode="pihcore.rad_order.title">
+      <section id="radiology-ultrasound-orders" sectionTag="fieldset" headerTag="legend" headerStyle="title" headerCode="pihcore.order.title">
+        <h3>Radiology and ultrasound orders</h3>
+        <div class="section-container-color">
+            <p>
+                <h3>X-ray 1 ordered</h3>
+                <select name="x-ray1_ordered">
```
lnball: These aren't htmlform tags.  Can you update this @lebbiesahr ?

lebbiesahr: yes, this code was added to let you see what i wanted to put in - i don't know if i should add these lists for x-ray/ultrasound as they are i.e. local to SL implementation or try to make part of wider list.  I think they are specific sub-lists so could keep local.  Will amend the tags once approach agreed.

lnball: @lebbiesahr  I would recommend that you add "real" code but comment it out until the concepts are available.  For example,

omod/src/main/webapp/resources/htmlforms/sierra_leone/section-exam.xml
```diff
@@ -740,7 +763,7 @@
 
         <p/>
 
-        <div class="section-container-color">
+     <!-->   <div class="section-container-color">
```
lnball: This might work, but html comment format is normally this:

`<!--  fskfjklsdjflkjf fkjfksdjfklsdjflksjfl -->`

not
`<!-->`

omod/src/main/webapp/resources/htmlforms/sierra_leone/section-plan.xml
```diff
@@ -179,288 +179,63 @@
...
-                         style="checkbox"
-                         showCommentField="true" commentFieldLabel=""/>
-                </p>
-            </div>
+            <p>
+                <obs conceptId="PIH:FAMILY PLANNING COMMENT" 
+                labelText="Offered and accepted Family Planning information"  
+                answers="PIH:YES,PIH:NO" 
+                answerLabels="Yes,No" 
+                style="dropdown"
```
lnball: I would recommend avoiding dropdowns -- it requires an extra click.  I've been moving to radio buttons for all of these. @lebbiesahr

lebbiesahr:
@lnball - have made some amendments according to your suggestions.  If you could approve that would be good.

PR #96, opened 2019-08-06
MEX-404 Registration form elements should be in Spanish

brandones:
This adds names to the MDD concepts. I did this before I found out that
those names don't seem to do anything -- the names used in the reg form
are in messages.properties.


mogoodrich:
LGTM

PR #97, opened 2019-08-06
Fix duplicate name Ocupación

brandones:
For https://pihemr.atlassian.net/browse/MEX-404

Fixes problem introduced in https://github.com/PIH/openmrs-module-pihcore/pull/96


brandones:
Trivial, merging.

PR #103, opened 2019-09-17
SL-120: Support country-specific configs in encounterTypeConfig (in o…

mogoodrich:
…rder to add MCH to for Sierra Leone)

### DIFF
--- a/omod/src/main/webapp/resources/scripts/visit/encounterTypeConfig.js
+++ b/omod/src/main/webapp/resources/scripts/visit/encounterTypeConfig.js
@@ -100,31 +100,6 @@ angular.module("encounterTypeConfig", ["constants"])
             require: "patientAgeInYearsOnDate(visit.startDatetime) < 15"
         };
 
-        // ToDo: Remove this section.  It was combined with supplements
-        /*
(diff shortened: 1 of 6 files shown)


omod/src/main/webapp/resources/scripts/visit/encounterTypeConfig.js
```diff
@@ -305,8 +280,40 @@ angular.module("encounterTypeConfig", ["constants"])
...
+         *      DEFAULT: {
+         *        // specific country config
+         *      }
+         *      SPECIFIC_SITE: {
+         *        // specific site config
+         *      }
+         *    }
+         * }
+         **/
+
```
mogoodrich: I had hoped to support having a country or site "automatically" pick up the forms from the proper site or country sub-directory, but I realized this isn't (easily) possible because all this config is happening on the client side so we'd have to query the server to see if a country or site-specific form was available and otherwise fall back to the default.

So configuring country and site-specific forms will still require editing the encounterTypeConfig as shown above... however, we probably want to support completely different configs per country anyway, and with this change, we can do that without defining a new encounter type.

PR #105, opened 2019-09-18
Uhm 3712 social economics report

jmbabazi:


api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -59,6 +59,7 @@
         public static final String PRIMARY_CARE = "primaryCare";  // primary care forms for Haiti, Mexico, and Sierra Leone (country-dependent)
         public static final String PACS_INTEGRATION = "pacsIntegration";
         public static final String ED_TRIAGE = "edTriage";
+        public static final String SOCIAL_ECONOMICS = "socialEconomics";
```
mogoodrich: I think this line can be removed, as we don't need a new component for this (see my comments on the Mirebalais Reports PR)

jmbabazi: Thanks Mark. Removed

jmbabazi:
@mogoodrich @dmdesimone , I'm merging this in. We had forgotten to merge them earlier

PR #107, opened 2019-09-25
UHM-4083 Order encounter for labs using pathology display template

Jilouwikyky:


api/src/main/resources/messages_es.properties
```diff
@@ -859,3 +859,5 @@ pihcore.unknown.oneChar=u
 pihcore.unknown=unknown
 
 pihcore.details=Details
+
+pihcore.test=Test
```
mogoodrich: some test comment

PR #109, opened 2019-09-26
MEX-459 "Confirmed" checkbox in Diagnosis wiget should instead say "First time" ("1a vez")

brandones:
https://pihemr.atlassian.net/browse/MEX-459

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
+++ b/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
@@ -1046,6 +1046,24 @@
         </h1>
     </ifMode>
 
+    <script type="text/javascript">
+        // Hack to make "Confirmado" appear as "1a vez"
+        // See https://pihemr.atlassian.net/browse/MEX-459
+
+        $('body').on('DOMNodeInserted', 'div', function () {
+            const targetText = "1a vez";
+            const confirmedLabel = jq(jq('input[data-ng-model="d.confirmed"]')[0]).parent();
+            const confirmedText = $(confirmedLabel[0])
+                    .contents()
+                    .filter(function () {
+                        return this.nodeType == 3;  // this picks out the text content of the label
+                    })[1];
+            if(confirmedText !== undefined) {
+                confirmedText.textContent = "  " + targetText;  // add a couple spaces margin
+            }
+        });
+    </script>
+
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -1046,6 +1046,24 @@
         </h1>
     </ifMode>
 
+    <script type="text/javascript">
+        // Hack to make "Confirmado" appear as "1a vez"
+        // See https://pihemr.atlassian.net/browse/MEX-459
+
+        $('body').on('DOMNodeInserted', 'div', function () {
+            const targetText = "1a vez";
+            const confirmedLabel = jq('label:contains("Confirmado")');
```
mogoodrich: Is there an id on this label?  Seems like there's a risk here if someone changes the display test to something other than "Confirmado", which I would think someone might assume they could do without causing any downstream changes?

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -1046,6 +1046,24 @@
...
+            const confirmedText = $(confirmedLabel[0])
+                    .contents()
+                    .filter(function () {
+                        return this.nodeType == 3;
+                    })[1];
```
mogoodrich: Not quite sure what this is doing, but I assume it works.

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -1046,6 +1046,24 @@
         </h1>
     </ifMode>
 
+    <script type="text/javascript">
+        // Hack to make "Confirmado" appear as "1a vez"
+        // See https://pihemr.atlassian.net/browse/MEX-459
+
+        $('body').on('DOMNodeInserted', 'div', function () {
+            const targetText = "1a vez";
+            const confirmedLabel = jq(jq('input[data-ng-model="d.confirmed"]')[0]).parent();
```
brandones: @mogoodrich Good thought about the lookup by label. I've changed it to refer to a data property of the <input> tag it contains.

brandones:
Thanks for the review @mogoodrich !

PR #113, opened 2019-10-07
MEX-409 Users should be able to print prescriptions for patients

brandones:
https://pihemr.atlassian.net/browse/MEX-409


cioan:
Looks good @brandones ! Thanks.

PR #121, opened 2019-10-30
MEX-402 Patient search fuzzy matcher should work for Spanish

brandones:
https://pihemr.atlassian.net/browse/MEX-402

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/search/PihLuceneAnalyzerFactory.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/search/PihLuceneAnalyzerFactory.java
@@ -1,0 +1,43 @@
+package org.openmrs.module.pihcore.search;
+
+import org.apache.lucene.analysis.core.KeywordTokenizerFactory;
+import org.apache.lucene.analysis.core.LowerCaseFilterFactory;
+import org.apache.lucene.analysis.core.WhitespaceTokenizerFactory;
+import org.apache.lucene.analysis.miscellaneous.ASCIIFoldingFilterFactory;
+import org.apache.lucene.analysis.ngram.EdgeNGramFilterFactory;
+import org.apache.lucene.analysis.ngram.NGramFilterFactory;
+import org.apache.lucene.analysis.standard.ClassicFilterFactory;
+import org.hibernate.search.annotations.Factory;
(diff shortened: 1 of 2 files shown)


mogoodrich:
Seems cool and simple to do if we get it to work! :)

api/src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,22 @@
...
+-->
+<!DOCTYPE hibernate-configuration PUBLIC
+        "-//Hibernate/Hibernate Configuration DTD 3.0//EN"
+        "http://www.hibernate.org/dtd/hibernate-configuration-3.0.dtd">
+
+<hibernate-configuration>
+
+    <session-factory>
+        <property name="hibernate.search.model_mapping">org.openmrs.module.pihcore.search.PihLuceneAnalyzerFactory</property>
+    </session-factory>
```
mseaton: I don't know that a module can override the hibernate.cfg.xml file like this.  Do you know that this is possible?

api/src/main/java/org/openmrs/module/pihcore/search/PihLuceneAnalyzerFactory.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.lucene.analysis.standard.ClassicFilterFactory;
+import org.hibernate.search.annotations.Factory;
+import org.hibernate.search.cfg.SearchMapping;
+import org.openmrs.api.db.hibernate.search.LuceneAnalyzers;
+
+
+public class PihLuceneAnalyzerFactory {
+    @Factory
+    public SearchMapping getSearchMapping() {
+        SearchMapping mapping = new SearchMapping();
```
mseaton: One approach would be for this SearchMapping object to be defined as a Spring bean, where there is a default bean supplied by core, and modules can override this by adding their own bean that implements the same interface and the code within here chooses one vs. the other based on some configured sort order.   There are examples of doing this within the reporting module and elsewhere that I could point you to.  Not sure if that is the most efficient model or not, but it's an option.

mogoodrich: @mseaton @brandones this seems like a good approach to me!

PR #125, opened 2019-11-04
MEX-439 Show vitals from the current visit at the top of the consult note (revision)

brandones:
Revision: fix message-properties, add units, clean up

### DIFF
--- a/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
+++ b/omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
@@ -86,6 +86,10 @@
             min-width: initial;
         }
 
+        .light-font {
+            font-family: "OpenSansLight",sans-serif;
+        }
+
         .small {
             max-width: 10em;
             display: block;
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -193,156 +197,115 @@
...
                     <tr>
                         <td>
-                            <uimessage code="pihcore.vitals.systolic.bp.title"/>
+                            <uimessage code="mirebalais.vitals.bloodPressure.title" />
                         </td>
                         <td>
-                            <p>
-                                <lookup expression="fn.latestObs('PIH:SYSTOLIC BLOOD PRESSURE').obsDatetime"/>
-                            </p>
+                            <lookup expression="fn.latestObs('PIH:SYSTOLIC BLOOD PRESSURE').obsDatetime"/>
```
brandones: Assume that diastolic and systolic will always be recorded together.

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -193,156 +197,115 @@
...
-                                <lookup expression="fn.latestObs('PIH:DIASTOLIC BLOOD PRESSURE').obsDatetime"/>
-                            </p>
-                        </td>
-                        <td>
-                            <p>
-                                <lookup expression="fn.latestObs('PIH:DIASTOLIC BLOOD PRESSURE').valueNumeric"/>
-                            </p>
+                            <lookup expression="fn.latestObs('PIH:SYSTOLIC BLOOD PRESSURE').valueNumeric.intValue()"/>
+                            <lookup complexExpression="#if($fn.latestObs('PIH:SYSTOLIC BLOOD PRESSURE')) / #end" />
+                            <lookup expression="fn.latestObs('PIH:DIASTOLIC BLOOD PRESSURE').valueNumeric.intValue()"/>
```
brandones: Format like "120 / 80"

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -193,156 +197,115 @@
...
-                            </p>
+                            <lookup expression="fn.latestObs('PIH:SERUM GLUCOSE').obsDatetime"/>
                         </td>
                         <td>
-                            <p>
-                                <lookup expression="fn.latestObs('PIH:SERUM GLUCOSE').valueNumeric"/>
-                            </p>
+                            <lookup expression="fn.latestObs('PIH:SERUM GLUCOSE').valueNumeric.intValue()"/>
+                            <span class="light-font">
+                                <lookup complexExpression="#if($fn.latestObs('PIH:Fasting for blood glucose test')) #if($fn.latestObs('PIH:Fasting for blood glucose test').valueCoded.name != 'No') (en ayuno) #else (sin ayuno) #end #end" />
```
brandones: Unfortunately hard-codes some things that should be message properties ("en ayuno", "sin ayuno"), but oh well. Velocity.

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
```diff
@@ -193,156 +197,115 @@
...
-                            <uimessage code="pihcore.lab.isFastingGlucose"/>
-                        </td>
-                        <td>
-                            <p>
-                                <lookup expression="fn.latestObs('PIH:Fasting for blood glucose test').obsDatetime"/>
-                            </p>
+                            <lookup expression="fn.latestObs('PIH:BLOOD OXYGEN SATURATION').obsDatetime"/>
                         </td>
                         <td>
-                            <p>
```
brandones: `<p>` tags not needed within `<td>` -- here they just create some additional vertical spacing, which I don't think is needed.

omod/src/main/webapp/resources/htmlforms/mexico/consult.xml
jmbabazi: @brandones , I had done this approach before but then I realized the info might be misleading in cases where they are not done together. maybe there could be another approach where it evaluates and ensures that the obs_datetime is the same?

brandones: Yeah, while that would be great, it would be an extreme corner case -- it would mean that someone screwed up data entry pretty badly. But you're right, it is misleading in the case of that error. Given the constraints of our tools, I think we have to be okay with that.

PR #126, opened 2019-11-05
LIB - Epilepsy section within the Mental Health form; removed 'include' feature to be implemented within HFE

alaboso:
- Epilepsy section within the Mental Health form; 
- Removed 'include' feature to be implemented within HFE


api/src/main/resources/messages.properties
```diff
@@ -891,3 +891,20 @@ pihcore.medication.side.effects=Side Effects
 pihcore.test=Test
 
 pihcore.vitals.sierraLeone.confused=Confused
+
+pihcore.epilepsy.care.during.seizure=Care during seizure?:
+pihcore.epilepsy.recognized.trigger.signs=Recogniz. Trigg. &amp; signs?:
+pihcore.epilepsy.habits.duration=Duration:
```
lnball: Duration and some of these are already in the messages.properties.  This is fine @alaboso , but in the future just use the message that's already in the file.  

We shouldn't continue a habit of including the specifics of where the string is used in the messages files (ie. pihcore.visitNote.plan.duration).

PR #127, opened 2019-11-07
added hiv counselling to plan, moved chorioaminiotisis in MCH delivery

lebbiesahr:
@mogoodrich/ @lnball  request adds hiv counselling tick box and transfers a finding in MCH delivery form


lnball:
I am merging this in with my code since it would conflict.  (The id's for the HIV counseliing and Family Planning sections are flipped.)

PR #136, opened 2019-12-23
Add SPA component

brandones:


api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -76,4 +76,5 @@
         public static final String ORDER_ENTRY = "orderEntry";  // TODO we probably want a different name for this?  break up by drug orders and lab orders, etc?
         public static final String COHORT_BUILDER = "cohortBuilder";
         public static final String CHEMOTHERAPY = "chemotherapy";
+        public static final String SPA = "spa";  // the Single-SPA UI
```
mogoodrich: For what it's worth, once we confirm this is working fine we should just remove this component and use the spa homepage everywhere, this component shouldn't have to hang around for too long.

PR #137, opened 2020-01-09
PER-334: Set up site for Peru

brandones:
https://pihemr.atlassian.net/browse/PER-334


api/src/main/java/org/openmrs/module/pihcore/identifier/peru/ConfigurePeruIdGenerators.java
```diff
@@ -0,0 +1,59 @@
...
+import org.openmrs.Location;
+import org.openmrs.PatientIdentifierType;
+import org.openmrs.module.idgen.AutoGenerationOption;
+import org.openmrs.module.idgen.SequentialIdentifierGenerator;
+import org.openmrs.module.idgen.service.IdentifierSourceService;
+import org.openmrs.module.metadatadeploy.MetadataUtils;
+import org.openmrs.module.pihcore.config.Config;
+import org.openmrs.module.pihcore.metadata.peru.PeruPatientIdentifierTypes;
+
+public class ConfigurePeruIdGenerators {
```
mogoodrich: tangentially, we should make sure we remember to let the SES team know that this is just a default format we came up with, and there are a lot of custom options available we could use.

brandones: I'm going to swap this out with a CSV as soon as I implement the Iniz PIT loader

PR #141, opened 2020-01-29
UHM-4257: Move pih-config files out of Mirebalais Puppet into new "op…

mogoodrich:
…enmrs-config-pih-[country]" repos

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/config/ConfigLoader.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/config/ConfigLoader.java
@@ -1,7 +1,6 @@
 package org.openmrs.module.pihcore.config;
 
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.codehaus.jackson.JsonNode;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/pihcore/config/ConfigLoader.java
```diff
@@ -42,7 +43,7 @@ public static String getPihConfigurationDirRuntimeProperty(String defaultValue)
      * Loads Configuration based on configuration in the runtime properties file
      */
     public static ConfigDescriptor loadFromRuntimeProperties() {
-        return load(getRuntimeConfiguration("pihcore"));
+        return load(getRuntimeConfiguration("default,site-default"));
```
mogoodrich: By default, if no "pih.config" runtime-property is found, OpenMRS will configure using the following two config files:

configuration/pih/pih-config-default.json
configuration/pih/pin-config-site-default.json

This allows someone to fire up a base configuration without the hacky step to stop and update their runtime.properties file.

I'm setting up these files via sym links in the various project directories.

PR #143, opened 2020-02-07
LIB - Added Community health role and updated ncd initial forms

alaboso:
- Added COMMUNITY_HEALTH_WORKER role and enabled more Roles in the Liberian deployments

- Updated ncd-initial CHW person role to be community health worker and added i18n property for community health worker privileges


lnball:
@alaboso Do you expect that CHWs will have user accounts for the EMR?  The code looks good.

alaboso:
Yes CHWs will have user accounts. It may not happen immediately but yes, they will.

PR #144, opened 2020-02-07
UHM-4208: Move forms from PIH Core to pih-config-pihemr

mogoodrich:


brandones:
@mogoodrich I'll wait to review this until after you've resolved the conflicts?

mogoodrich:
@brandones I will try to resolve the encounterTypeConfig.js conflicts today, but the other three are forms that I deleted and assumedly have been changed since then... so for these although it's critical that I actually copy over any changes to the forms to the new config-repos and make sure everything is sync, it's a moving target and something I need to do last-minute before the merge, and shouldn't affect your ability to review?

PR #146, opened 2020-02-13
SQL fucntion to return the Latest Observation

jmbabazi:


mseaton:
@jmbabazi - I committed this in a new commit as it was easier that way to resolve the conflict.  Thanks!

PR #148, opened 2020-02-25
UHM-4721

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -115,7 +131,10 @@ public void stopped() {
     public void contextRefreshed() {
 
         setDispositionConfig(config);
+        setMetadataSharingResolvers();
```
brandones: If this is all the programmatic MDS config we do, can we call this function `configureMetadataSharing()`?

api/pom.xml
```diff
@@ -39,9 +39,10 @@
 		<testResources>
 			<testResource>
 				<directory>src/test/resources</directory>
-				<filtering>true</filtering>
+				<filtering>false</filtering>
```
brandones: What does this do?

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataSharingSetup.java
```diff
@@ -0,0 +1,150 @@
...
+
+    protected static final Log log = LogFactory.getLog(MetadataSharingSetup.class);
+
+    public static void installMetadataSharingPackages() {
+
+        try {
+            Collection<File> mdsFiles = loadMdsFiles();
+
+            if (mdsFiles == null) {
+                log.error("Unable to find any metadata sharing files");
```
brandones: This is redundant with the error messages in `loadMdsFiles`

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
brandones: Actually, should this go in `MetadataSharingSetup`, and just be handled internally by that class?

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataSharingSetup.java
```diff
@@ -0,0 +1,150 @@
...
+                return;
+            }
+
+            List<PackageImporter> packageImporters = loadPackageImporters(mdsFiles);
+            removeAlreadyLoadedPackageImporters(packageImporters);
+            sortPackageImporters(packageImporters);
+            loadPackages(packageImporters);
```
brandones: All this logic seems quite a lot more complicated than what it replaces, which is

```
MetadataUtil.setupSpecificMetadata(getClass().getClassLoader(), "some", "package", "names")
```

Why aren't we doing it that way, with emrapi?

api/pom.xml
mogoodrich: We added some test zip files into the test directory, and Maven was doing variable-replacement filtering on these files, which was corrupting them.

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mogoodrich: Agreed, this method should live in MetadataSharingSetup.  I'm going to keep it as a separate method within that class, because we want to call it from the contextRefreshed methoid, not the started method.

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataSharingSetup.java
mogoodrich: Good point, removed.

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataSharingSetup.java
mogoodrich: Yeah, I had been hoping/assuming we'd just us the EmrApi methods, but I ran into a few hiccups.  First, the EmrApi functionality was based around loading the packages from the classpath, but, more, importantly, the EmrApi method depends on the "packages.xml" external file to determine what version a package represents and what the groupUuid of that package is.

The new way loads all the files it finds in the concept directory and then retrieves the groupUuid  and the version number by parsing the package itself. (I though this might be time-consuming, but it turned out not to be).

So I was annoyingly unable to use the EmrApi code directly.  Since this is likely a shorter term fix I just copied and reworked code as necessary from EmrApi into PIH Core rather than updating EmrApi to support these changes (it's also likely that no one other that us uses these EmrApi uitl methods).

PR #149, opened 2020-02-27
CCI-43, add commcare user

cioan:
@mseaton and @mogoodrich , here is the PR for creating a CommCare user with credentials specified in the openmrs-runtime.properties. Does this make sense to you? Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
@@ -17,7 +17,12 @@ package org.openmrs.module.pihcore;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.openmrs.*;
+import org.openmrs.GlobalProperty;
+import org.openmrs.PatientIdentifierType;
+import org.openmrs.User;
+import org.openmrs.Person;
+import org.openmrs.PersonName;
+import org.openmrs.Provider;
 import org.openmrs.api.AdministrationService;
 import org.openmrs.api.ConceptService;
 import org.openmrs.api.EncounterService;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -14,10 +14,10 @@
 package org.openmrs.module.pihcore;
 
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.openmrs.GlobalProperty;
-import org.openmrs.PatientIdentifierType;
+import org.openmrs.*;
```
mogoodrich: A nitpick, but we shouldn't be using global imports as it's OpenMRS convention not to... do you have IntelliJ doing this replacement?

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -220,6 +222,48 @@ public void setGlobalProperties(Config config) {
         }
     }
 
+    public User setupCommCareUser(){
+
+	    User user = null;
+        String commCareUserName = Context.getRuntimeProperties().getProperty("commcare.username", "commCareUser");
+        String commCareUserPassword = Context.getRuntimeProperties().getProperty("commcare.password", "CommCare12");
```
mogoodrich: We should make the default values here blank and not install a user if the values aren't specified in the runtime properties... otherwise we could inadvertently create a user with the username/password pair in plaintext in a public repo... :)  Then we'd only set these runtime properties in Puppet for the servers that need this user.

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -220,6 +222,48 @@ public void setGlobalProperties(Config config) {
...
+            user.setUsername(commCareUserName);
+            Person person = user.getPerson();
+            if (person == null) {
+                person = new Person();
+                person.addName(new PersonName("CommCare", "", "Integration"));
+                person.setGender("M");
+                user.setPerson(person);
+            }
+
+            user.addRole(Context.getUserService().getRole("Authenticated"));
```
mogoodrich: What's this role, I forget? (Though from the name it does seem correct that this user should have it... :) )

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -220,6 +222,48 @@ public void setGlobalProperties(Config config) {
...
+                user = Context.getUserService().createUser(user, commCareUserPassword);
+            } else {
+                // update an existing record
+                user = Context.getUserService().saveUser(user);
+            }
+            person = user.getPerson();
+            Collection<Provider> providers = Context.getProviderService().getProvidersByPerson(person);
+            if (providers == null || providers.size() < 1 ){
+                // if the user does not have a Provider yet then create one
+                Provider provider = new Provider();
```
mogoodrich: I *believe* that we will need to hard-code the uuid of this provider in the CommCare integration (so we would need assign a hardcoded uuid here?).

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
cioan: I think that basically this gives read permissions:
https://humci.pih-emr.org/mirebalais/admin/users/role.form?roleName=Authenticated

cioan:
Thanks @mogoodrich ! I have addressed all your comments above. Thanks.

cioan:
Thanks @mogoodrich !

PR #151, opened 2020-03-02
Bump pihcore Java version to 8

brandones:
Everything seems to be working on my local server, and tests are passing.

https://pihemr.atlassian.net/browse/UHM-4411


mogoodrich:
LGTM, though let's either merge it in this evening when we have time to watch the build and confirm that everything works, or wait until after the retreat (so if something does go wrong, we don't have our servers down during the retreat).

PR #163, opened 2020-04-08
UHM-4459: Load Concepts in Background

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -110,13 +115,21 @@ public void started() {
...
-            DrugListSetup.installDrugList();
             PacIntegrationSetup.setup(config);
             AttachmentsSetup.migrateAttachmentsConceptsIfNecessary(conceptService);
            // RetireProvidersSetup.setupRetireProvidersTask();
             setupCommCareUser();
+
+            // see https://pihemr.atlassian.net/browse/UHM-4459 for details of why we do this this way
+            Runnable metadataSetupTask = new MetadataSetupTask(config, disableInstallMetadataBundlesThatDependOnMDSPackages);
+            String runInSeparateThread = Context.getAdministrationService().getGlobalProperty(PihCoreConstants.GP_RUN_CONCEPT_SETUP_TASK_IN_SEPARATE_THREAD);
+            if (StringUtils.isNotBlank(runInSeparateThread) && runInSeparateThread.toLowerCase().equals("true")) {
```
mseaton, line 126: could just say `if ("true".equalsIgnoreCase(runInSeparateThread))`

api/src/main/java/org/openmrs/module/pihcore/setup/HtmlFormSetup.java
```diff
@@ -20,4 +29,26 @@ public static void setupHtmlFormEntryTagHandlers() throws Exception {
...
+            for (File file : files) {
+                try {
+                    HtmlFormUtil.getHtmlFormFromUiResource(resourceFactory, formService, htmlFormEntryService, "file:" + file.getAbsolutePath());
+                } catch (IOException e) {
+                    log.error("Unable to load HTML Form at path: " + file, e);
+                }
+            }
+        }
+        catch (Exception e) {
+            log.error("Unable to load HTML forms--this error is expected when running component tests");
```
mseaton, line 51: If it is not expected when you are _not_ running component tests, don't we want to throw the exception here?  Or maybe log the error and return a boolean indicating whether it succeeded or not so that calling code can determine how to handle this?

api/src/main/java/org/openmrs/module/pihcore/setup/HtmlFormSetup.java
```diff
@@ -20,4 +29,26 @@ public static void setupHtmlFormEntryTagHandlers() throws Exception {
...
+        try {
+            File htmlformDir = new File(PihCoreUtil.getFormDirectory());
+            Collection<File> files = FileUtils.listFiles(htmlformDir, null, true);
+
+            for (File file : files) {
+                try {
+                    HtmlFormUtil.getHtmlFormFromUiResource(resourceFactory, formService, htmlFormEntryService, "file:" + file.getAbsolutePath());
+                } catch (IOException e) {
+                    log.error("Unable to load HTML Form at path: " + file, e);
+                }
```
mseaton, line 47: Same comment here as below - why are we swallowing this error?

api/src/main/java/org/openmrs/module/pihcore/setup/HtmlFormSetup.java
```diff
@@ -20,4 +29,26 @@ public static void setupHtmlFormEntryTagHandlers() throws Exception {
...
+            }
+        }
+        catch (Exception e) {
+            log.error("Unable to load HTML forms--this error is expected when running component tests");
+        }
+    }
 }
```
mseaton, line 54: Given that this is part of the PR, maybe this is about more than just loading concepts in the background?

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataSetupTask.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            HtmlFormSetup.loadHtmlForms(); // this must happen *after* MDS packages are installed, so that forms defined in code/github take precedent
+            log.info("HTML forms loaded");
+        }
+        catch (Exception e) {
+            log.error("Aborting Metadata Setup Task: error installing HTML forms", e);
+            throw e;
+        }
+    }
+
```
mseaton, line 69: LGTM, but maybe just rename this PR / ticket to reflect it is more than just Concepts in Background

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mogoodrich: Good point, will change.

api/src/main/java/org/openmrs/module/pihcore/setup/HtmlFormSetup.java
mogoodrich: True... I was struggling with the name of the task, for what it's worth.  MetadataSetupTask seemed to vague, especially when it doesn't set up *all* metadata, but ConceptsAndDrugsAndFormsAndProgramsMetadataSetupTask didn't sound good as well.  In the end I went with MetadataSetupTask with the hopes that this will all be refactored within the year anyway.

PR #166, opened 2020-04-14
CCI 64 pediatric form

jmbabazi:


lnball:
Looks good, James.

messages.properties are not the cleanest.  In the future, you might check translations in another module (ie. mirebalais). mirebalais/api/src/main/resources/messages.properties has mirebalais.vitals.height.title = Height.  You can use these in the htmlforms and they are already translated.

PR #168, opened 2020-04-16
UHM-4501: Retire Liberia locations JJ Dossen and Pleebo and migrate data

mogoodrich:


api/src/main/resources/liquibase.xml
```diff
@@ -1116,5 +1116,30 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20200416-fix-liberia-locations" author="mgoodrich">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                select count(*) from location where uuid='f6f87d10-f290-11e4-b939-0800200c9a66';
```
mseaton, line 1122: Why limit to pleebo here, but include JJ Dossen in the update below?  Is JJ Dossen not something we need to check for here?

alaboso: This change doesn't affect JJDossen because the JJDossen initial rollout had this location structure wherease Pleebo had the old structure.

mogoodrich: Thanks @alaboso .  Also, for what it's worth, this will affect any server that has the Pleebo location *installed*, which I believe is JJ Dossen as well, so this will actually run on JJ Dossen (but it shouldn't actually change things there).  The main thing here is to make sure this doesn't run on any non-Liberia servers.

PR #169, opened 2020-04-22
UHM-4445: Migrate Haiti, Liberia, and Sierra Leone address bundles to…

mogoodrich:
… their respective config files


api/src/test/java/org/openmrs/module/pihcore/TestAddressBundle.java
```diff
@@ -1,46 +1,45 @@
-package org.openmrs.module.pihcore.deploy.bundle.sierraLeone;
+package org.openmrs.module.pihcore;
```
mogoodrich, line 1: This is a whole new file, copied (and edited) from Haiti Core, not sure why it's showing up as a diff.

api/src/test/java/org/openmrs/module/pihcore/reporting/BaseReportTest.java
```diff
@@ -67,7 +67,7 @@
     protected MetadataDeployService deployService;
 
     @Autowired
-    protected HaitiAddressBundle haitiAddressBundle;
+    protected TestAddressBundle testAddressBundle;
```
mogoodrich, line 70: Eventually we will want to set something up where we can initialize a test context using Iniz files, but here I just copied over the HaitiAddressBundle from Haiti Core, trimming the CSV to 5 lines, and renamed it "Test Address Bundle"

PR #177, opened 2020-06-03
Cleanup mostly related to UHM-4608

brandones:
EncounterTypeConfig should not reference constants library

https://pihemr.atlassian.net/browse/UHM-4608


brandones:
I searched for usages in js files in pihcore, mirebalais, and config-pihemr. Let me know if there are other places I should look.

omod/src/main/webapp/resources/scripts/visit/vaccinations.js
```diff
@@ -9,8 +9,8 @@ angular.module("vaccinations", [ "constants", "ngDialog", "obsService", "encount
```
brandones, line 9: In this file, really just taking cues from VSCode dimming unused variables.

omod/src/main/webapp/resources/scripts/visit/vaccinations.js
```diff
@@ -9,8 +9,8 @@ angular.module("vaccinations", [ "constants", "ngDialog", "obsService", "encount
 
     })
 
-    .factory("VaccinationService", [ "Obs", "Concepts", "EncounterTypes", "EncounterTransaction", "EncounterRoles", "SessionInfo",
-        function(Obs, Concepts, EncounterTypes, EncounterTransaction, EncounterRoles, SessionInfo) {
```
mogoodrich, line 13: I take it turns out we were including these libraries but no longer using them?

mogoodrich:
Yes, you shouldn't have to search anywhere outside of PIH Core @brandones

omod/src/main/webapp/resources/scripts/visit/vaccinations.js
brandones: Correct

PR #178, opened 2020-06-04
Adding a few functions to retrieve person attributes and encounter lo…

mseaton:
…cations

These are a few functions that seemed useful as I was creating the covid19 encounter export.  @dmdesimone / @jmbabazi do these look ok to you?  @mogoodrich FYI


brandones:
These are great! Should we start documenting the available ones in the pihcore README?

PR #184, opened 2020-06-14
Update create_functions.sql

dmdesimone:
For, UHM-4651, create a function to return primary identifier for a patient.

### DIFF
--- a/api/src/main/resources/sql/procedures/create_functions.sql
+++ b/api/src/main/resources/sql/procedures/create_functions.sql
@@ -172,6 +172,36 @@ RETURN dosId_out;
 
 END;
 #
+
+/*
+The following will return the primary patient identifier defined in the system's global properties for a given patient id 
+*/
+#
+DROP FUNCTION IF EXISTS primary_patient_id;
+#
+CREATE FUNCTION primary_patient_id(
+    _patient_id int
+)
+	RETURNS VARCHAR(255)
+    DETERMINISTIC
+
+BEGIN
+    DECLARE  id VARCHAR(255);
+
+    SELECT pid.identifier into id from patient_identifier pid 
+    INNER JOIN metadatamapping_metadata_term_mapping m on m.code='emr.primaryIdentifierType'
+    where voided = 0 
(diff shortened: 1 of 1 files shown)


mseaton:
@dmdesimone - I have another pending change to this file to look up global properties and user properties.  I think we should also add a function to get metadata mappings, which is currently embedded in this particular function.  Let me look at adding that in before we merge this...

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -172,6 +172,36 @@ RETURN dosId_out;
 
 END;
 #
+
+/*
+The following will return the primary patient identifier defined in the system's global properties for a given patient id
```
mogoodrich, line 177: change to "defined in the system's metadata mappings"... sorry I led you astray with the GP solution but that has been deprecated.

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -172,6 +172,36 @@ RETURN dosId_out;
...
+
+BEGIN
+    DECLARE  id VARCHAR(255);
+
+    SELECT pid.identifier into id from patient_identifier pid 
+    INNER JOIN metadatamapping_metadata_term_mapping m on m.code='emr.primaryIdentifierType'
+    where voided = 0 
+    and pid.identifier_type = 
+      (select pid2.patient_identifier_type_id from patient_identifier_type pid2 where
+      pid2.uuid = m.metadata_uuid)
```
mogoodrich, line 196: Not sure why we are doing the join in metadatamapping_metadata_term_mapping... can't we just do something like ...pid2.uuid = (select uuid from  metadatamapping_metadata_term_mapping where code='emr.primaryIdentifierType')

mogoodrich:
@mseaton @dmdesimone I noticed this PR is still open?  Did we resolve this?  FWIW, create_functions has been moved to config-pihemr

brandones:
Seems like this should be re-opened against config-pihemr if needed.

PR #185, opened 2020-06-17
UHM-3614: Add logic to manage actions on visit page

mogoodrich:


omod/src/main/webapp/resources/scripts/visit/filters.js
```diff
@@ -479,6 +479,10 @@ angular.module("filters", [ "uicommons.filters", "constants", "encounterTypeConf
                 });
             };
 
+            var hasMemberThatEvaluatesTrue = function(list, func) {
+              return list.some(func);
+            };
```
brandones, line 484: :upside_down_face:

brandones: If for some structural reason `hasMemberThatEvaluatesTrue` actually needs to exist, can we just call it `some`?

omod/src/main/webapp/resources/scripts/visit/filters.js
```diff
@@ -494,12 +498,12 @@ angular.module("filters", [ "uicommons.filters", "constants", "encounterTypeConf
 
                 if (it.require) {
                     // find a cleaner way to avoid polluting scope
-                    var result = (function(expr) {
+                    var result = (function(expr, visit) {
                         var sessionLocation = SessionInfo.get().sessionLocation;
                         var user = new OpenMRS.UserModel(SessionInfo.get().user);
                         var patient = visit ? visit.patient : null;
```
brandones, line 504: It appears that `visit` is already present in the closure. Why are you passing it as a parameter?

brandones:
Someone else is going to need to look at this one, this is unfamiliar territory.

PR #188, opened 2020-06-18
functions for valuedatetime and single questions that return yes

jmbabazi:


api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -847,3 +847,63 @@ BEGIN
...
+#
+
+-- This function accepts encounter_id, mapping source, mapping code (for both the concept_id and value_coded) and returns
+-- true if the obs_id exists and
+-- null if the obs_id does not exisit
+-- This function is used on questions that also act as answers (i.e you either check it as true or your don't)
+#
+DROP FUNCTION IF EXISTS obs_single_value_coded;
+#
+CREATE FUNCTION obs_single_value_coded(_encounterId int(11), _source varchar(50), _term varchar(255), _source1 varchar(50), _term1 varchar(255))
```
mseaton, line 885: I don't love these argument names.  Can you call these _questionSource, _questionTerm, _answerSource, _answerTerm, to make it clear what they are for?

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -847,3 +847,63 @@ BEGIN
...
+RETURNS varchar(11)
```
mseaton, line 886: I really do not think that this should return a varchar(11) with either null or "Yes".  That's a very specfic use case that is not super generalizable (and not localized, etc).  It would seem to make more sense if this function returned a boolean true/false (which are ultimately represented by 1/0 numerically, but which can be treated as booleans).  By having this function return true/false (or 1/0), you will be able to nest this function easily with others to do things like:

select if(obs_single_value_coded(...), 'YES', 'NO')

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -847,3 +847,63 @@ BEGIN
...
+DETERMINISTIC
+
+BEGIN
+
+DECLARE ret varchar(11);
```
mseaton, line 891: Per above, make this a "bit" datatype

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -847,3 +847,63 @@ BEGIN
...
+
+select      IFNULL(NULL, "Yes") into ret FROM
```
mseaton, line 893: Does this work? What is it checking?  Per my comments above, I'd change this to something like:

```
select      if(count(obs_id) > 0, 1, 0) into ret
from        obs o
where       o.voided = 0
and         o.encounter_id = _encounterId
and         o.concept_id = concept_from_mapping(_questionSource, _questionTerm)
and         o.value_coded = concept_from_mapping(_answerSource, _answerTerm)
```

If you are really trying to check that the most recent value in the encounter matches this, then use the order by and limit criteria that you added, but usually the common case is checking whether this value for this question exists (and multiple values might exist with check boxes, so you don't want to liimt to the most recent in the encounter).

PR #189, opened 2020-06-24
UHM-4383: Support setting up Location Tags in the pih-config files

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -111,6 +111,8 @@ public void started() {
             setGlobalProperties(config);
             setExtraIdentifierTypes(metadataMappingService, patientService, config);
             MergeActionsSetup.registerMergeActions();
+            // TODO line below will be removed when we migrate location tags
+            LocationTagSetup.setupLocationTagsLegacy(locationService, config);
             LocationTagSetup.setupLocationTags(locationService, config);
```
mogoodrich, line 116: We actually can start overriding what's defined in the "legacy" code with a config-based assignment immediately since this is called after the legacy.

It also occurs to me, fwiw, we can start migrating the location tag mapping prior to migrating the locations themselves to Iniz.  MIght be worth doing.

api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
```diff
@@ -16,17 +16,66 @@
...
-    // TODO actually, when we get rid of these, we can probably set most of the tags in the bundles?
+    private static final Logger log = LoggerFactory.getLogger(LocationTagSetup.class);
+
     public static void setupLocationTags(LocationService locationService, Config config) {
+        for (String locationTag : config.getLocationTags().keySet()) {
+            setLocationTagsFor(locationService, locationTag, config.getLocationTags().get(locationTag));
+        }
+    }
+
+    private static void setLocationTagsFor(LocationService service, String locationTag, List<String> locationsThatGetTag) {
```
mogoodrich, line 38: I took the existing setLocationTagsFor method and changed it to operate on a String instead of a LocationTagDescriptor and  a List<String> instead of a List<LocationDescriptor>.  I also moved the method up to the top of the class, which may not have been the best thing in terms of reviewing because you can't see the proper diff.

api/src/main/java/org/openmrs/module/pihcore/setup/LocationTagSetup.java
```diff
@@ -16,17 +16,66 @@
...
+                candidate.removeTag(tag);
+                service.saveLocation(candidate);
+            } else if (!actual && expected) {
+                candidate.addTag(tag);
+                service.saveLocation(candidate);
+            }
+        }
+    }
+
+    private static void setLocationTagsFor(LocationService service, LocationTagDescriptor locationTag, Collection<LocationDescriptor> locationsThatGetTag) {
```
mogoodrich, line 69: Provides support for the old signature. This method, and everything after it, can be removed once we migrate to config-based location tagging.

PR #193, opened 2020-07-01
Return date and obs_group_id basing on offset value

jmbabazi:


api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
 END
 
 #
+
+-- This function accepts encounter_id, mapping source, mapping code, offset value
+-- It will find a single, best observation that matches this, and return the value_datetime
+-- usage obs_value_datetime_by_offset(encounter_id, 'CIEL or PIH', "1234 OR TEXT");
```
mseaton, line 913: usage is missing the _offset_value arg

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
 END
 
 #
+
+-- This function accepts encounter_id, mapping source, mapping code, offset value
+-- It will find a single, best observation that matches this, and return the value_datetime
```
mseaton, line 912: Maybe more description about where the offset fits in - what use case is this solving?  when would you choose to use this function?

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
...
+
+RETURN ret;
+
+END
+
+#
+
+
+-- This function accepts encounter_id, mapping source, mapping code
+-- It will find a single, best observation that matches this, group them by the obs_group_id and return the value_text
```
mseaton, line 943: These comments need to be updated to reflect what is happening.

mseaton: I really don't know what this function does by reading it, and the comments aren't helpful in understanding.  Please document here what this function is intending to do and when you'd choose to use it.

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
...
+-- The CIEL1 or PIH1 are should be mappings for a concept that returns a value_datetime 
+#
+DROP FUNCTION IF EXISTS obs_value_coded_list_by_valuedatetime;
+#
+CREATE FUNCTION obs_value_coded_list_by_valuedatetime(_encounterId int(11), 
+														_source varchar(50), 
+															_term varchar(255), 
+																_locale varchar(50), 
+																	_source1 varchar(50), 
+																		_term1 varchar(50),
```
mseaton: 1 is not descriptive.  If source1 and term1 are meant to be referring to a datetime obs, you should name these so that this is more obvious.

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
...
+
+    DECLARE ret text;
+    DECLARE obs_gid int;
+
+-- return obs_group_id basing on value_datetime    
+    SELECT obs_group_id INTO obs_gid 
+    FROM obs 
+    WHERE voided = 0
+    AND encounter_id = _encounterId
+    AND DATE(value_datetime) = OBS_VALUE_DATETIME_BY_OFFSET(_encounterId, _source1, _term1, _offset_value);
```
mseaton: I'm confused why you have a DATE function on one side and no DATE function on the other.  I would expect you to need to convert both of these to a DATE for comparison, if they are both starting as datetime

api/src/main/resources/sql/procedures/create_functions.sql
```diff
@@ -907,3 +907,75 @@ RETURN ret;
...
+
+    SELECT      GROUP_CONCAT(DISTINCT CONCEPT_NAME(o.value_coded, _locale) SEPARATOR ' | ') INTO ret
+    FROM        obs o
+    WHERE       o.voided = 0
+      AND       o.encounter_id = _encounterId
+      AND       o.concept_id = CONCEPT_FROM_MAPPING(_source, _term)
+      AND 		obs_group_id = obs_gid;
+
```
mseaton, line 972: I would think that you are better off keeping this as 2 separate functions that you'd nest in your consuming report, rather than trying to create this larger function.  We should explore this.  For example, given 2 functions:

1. f1(_encounterId, _source1, _term1, _offset_value) -> returns an obs_group_id
2. f2(_encounterId, _source, _term, _obs_group_id) -> returns text of concept names

you'd use:

 f2(_encounter_id, _source, _term, f1(_encounter_id, _source1, _term1, _offsetValue))

This might lead to more reusable functions.  Assuming this performs equally well, might be a better choice.

jmbabazi:
Used other existing functions to solve this issue

PR #197, opened 2020-07-21
Lib 123

BranfordTGbieor:


BranfordTGbieor:
Hey @mogoodrich please take a look when you get the time.

api/src/main/java/org/openmrs/module/pihcore/metadata/core/EncounterTypes.java
```diff
@@ -259,6 +259,12 @@
 		public String description() { return "Vaccination form only (not within another encounter)"; }
 	};
 
+	public static EncounterTypeDescriptor ANC_FORM  = new EncounterTypeDescriptor() {
+		public String uuid() { return "c4ed1a17-cb3d-11ea-b84d-3c6aa7c392cc"; }
+		public String name() { return "ANC Form"; }
+		public String description() { return "General form for capturing all ANC-related data for Liberia"; }
+	};
+
```
mogoodrich, line 267: @BranfordTGbieor  @lnball do you think we want to use the existing either the existing ANC_INTAKE or ANC_FOLLOWUP encounter type for this?

Also, @BranfordTGbieor  if we do keep this encounter type we may want to consider making a new "LiberiaEncounterTypes" class, unless we think this encounter type could be used outside of Liberia. (There is an existing MexicoEncounterTypes for a pattern to follow). Would be interested in @lnball 's though on that one as well...

Regardless, if we do add this encounter type, we will need to add an "install" line to EncounterTypeBundle, or a new LiberiaEncounterTypeBundle class.

lnball: Since this covers ANC Intake AND followup, it would be good to create a new encounter type.  I would suggest the name of "ANC" (remove FORM from the name).  

It's possible it would be used outside of Liberia, but I don't know of a case for this now or in the near future.  I'd recommend a LiberiaEncounterTypeBundle.  I believe, we've already implemented in Haiti, Mexico, SL with the 2 different encounter types (ANC Intake/Followup)

mogoodrich: Thanks @lnball .  So, yes, based on what Ellen says I'd:

* Create a LiberiaEncounterTypes class in the package org.openmrs.module.pihcore.metadata.liberia and move the new encounter type there
* Create a LiberiaEncounterTypeBundle class in  the pacakge org.openmrs.module.pihcore.deploy.bundle.liberia to install that new encounter type
* Then, add a reference to LiberiaEncounterTypeBundle in the @Requires sectiojn of LiberiaMetadataBundle... this just basically tells OpenMRS to install all the bundles in the "Requires" list when installing the LiberiaMetadataBundle

For an example, it's basically the same pattern we are using in MexicoEncounterTypes, MexicoEncounterTypeBundle, and MexicoMetadataBundle.

BranfordTGbieor: That'd be good @mogoodrich, the new EncounterType class for Liberia is splendid idea! 
@lnball , Thanks for your insights on the matter too.

BranfordTGbieor:
Thanks alot @mogoodrich !

PR #200, opened 2020-08-05
UHM-4683 Move Location Tag Setup out of PIH Core and into Config

brandones:
https://pihemr.atlassian.net/browse/UHM-4683


api/src/test/java/org/openmrs/module/pihcore/reporting/MockLocationSetup.java
```diff
@@ -0,0 +1,4 @@
+package org.openmrs.module.pihcore.reporting;
```
mogoodrich, line 1: Do we use this at all?

brandones: Oh whoops, detritus! Deleting, thanks

PR #203, opened 2020-09-14
UHM-3122 Removing HIV Program workflow

brandones:
https://pihemr.atlassian.net/browse/UHM-3122


lnball (approved):
Looks right for future installs ;)  I forget if we need to do something to retire/deprecate these on an existing installation.  Maybe you already tested this @brandones ?

mogoodrich:
Exactly @lnball  I was wondering the same thing, but, as you said, @brandones  may have already tested.  If not, just a matter of making sure that workflow is removed from the DB or retired after it runs.

brandones:
Yes, I already tested. The workflow gets retired. Thanks @lnball and @mogoodrich !

PR #208, opened 2020-10-07
UHM-4880, install PatientIdentifierType with UniquenessBehavior

cioan:
@mogoodrich , here are the changes I had to make in the pihcore to be able to set the unique property on the PatientIdentifierType. This goes with the metadatadeploy PR:
https://github.com/openmrs/openmrs-module-metadatadeploy/pull/12

Thanks!


mogoodrich (commented):
Let's see if Daniel has anything to say about the MDD PR... if that's good, I think this is good.

A little annoying that you'd have to explicitly include the "1_11" version here, but I don't see a way around it (and it's still probably cleaner than any other option for doing this)

PR #209, opened 2020-10-09
UHM-4881, configure national and fiscal IDs

cioan:
@mogoodrich , here is the commit for configuring National ID and Fiscal ID. Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/haiti/PihHaitiPatientIdentifierTypeBundle.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/haiti/PihHaitiPatientIdentifierTypeBundle.java
@@ -19,6 +19,8 @@ public class PihHaitiPatientIdentifierTypeBundle extends AbstractMetadataBundle1
         install(PihHaitiPatientIdentifierTypes.USER_ENTERED_REF_NUMBER);
         install(PihHaitiPatientIdentifierTypes.DENTAL_DOSSIER_NUMBER);
         install(PihHaitiPatientIdentifierTypes.HIV_DOSSIER_NUMBER);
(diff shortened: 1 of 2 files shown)


mogoodrich (commented):
@cioan  I believe these two identifiers are already being installed?  likely in the Haiti Core module?

Of course that would raise the question "is it okay to add the non_unique constraint for everyone using Haiti Core", but since I'm not sure if eSante+ every started using it I think we can just make the change there and then reassess later if needed.

cioan:
Thanks @mogoodrich ! So, Haiti core is an OpenMRS module, so potentially could be used by other implementations? Is that OK to make those IDs NON-UNIQUE for everyone? I do not mind moving this code from pihcore to haiticore. Please let me know.

cioan:
Sorry @mogoodrich , I just saw the second paragraph in your comment above. :-) I will move the code into the haiticore. Thanks!

mogoodrich:
Hmm... at first I was going to say go ahead and change it, but thinking it through a little more probably just safer to keep it in PIH Core...  what  you could actually do (I think) is change the descriptors in PIH Core to just extend the Haiti Core ones and then add the uniqueness behavior?

mogoodrich:
(The two identifier types would end up being installed twice, once in Haiti Core and once in PIH Core, but the PIH Core should happen second, so it should be okay)

cioan:
@mogoodrich , there was not an apparent way to me on how to extend HaitiPatientIdentifierTypes. Maybe because we have to use a different type, PatientIdentifierTypeDescriptor1_11? Anyway, I just retrieved the uuid, name and description from the HaitiPatientIdentifierTypes class. Thanks!

mogoodrich (approved):
Perfect, thanks @cioan !

PR #211, opened 2020-10-16
UHM-4931: allow adding sites to PIH EMR without needing to update Sit…

mogoodrich:
…e enum

https://pihemr.atlassian.net/browse/UHM-4931

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/config/ConfigDescriptor.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/config/ConfigDescriptor.java
@@ -19,13 +19,6 @@ public class ConfigDescriptor {
         HAITI, LIBERIA, SIERRA_LEONE, MEXICO, PERU, OTHER
     }
 
-    public enum Site {
-        MIREBALAIS, LACOLLINE, THOMONDE, BOUCANCARRE, ZLTRAINING, HINCHE, CERCA_LA_SOURCE, BELLADERE, HSN_SAINT_MARC,
-        PLEEBO, HARPER, OTHER, CROSS_SITE,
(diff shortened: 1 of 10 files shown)


brandones:
Seems pretty reasonable. Do you have a use case in mind?

brandones:
Ah, just read the ticket.

brandones (approved):

cioan (approved):
Looks good! Thanks @mogoodrich !

api/src/main/java/org/openmrs/module/pihcore/config/ConfigDescriptor.java
```diff
@@ -19,13 +19,6 @@
         HAITI, LIBERIA, SIERRA_LEONE, MEXICO, PERU, OTHER
     }
 
-    public enum Site {
```
mseaton, line 22: An alternative to this would have been to make Site an inner class, and then make the enum values into public String constants.  Then (I think) you could have left all of the code in place like ConfigDescriptor.Site.MIREBALAIS, as these would be String constants rather than enum values.  But it looks like this has been merged, so no worries either way @mogoodrich

mogoodrich: Yeah, I was thinking of something that left the variables in place as string constants, but it seemed like since (I think) we want to get away from the this kind of pattern in general (having direct references to specific sites within the code) I went with the  approach I did.

mseaton: Yep, definitely makes it more clear.

PR #213, opened 2020-10-30
Add workflows to HIV program

brandones:
https://pihemr.atlassian.net/browse/UHM-4977

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreConstants.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreConstants.java
@@ -97,7 +97,7 @@ public class PihCoreConstants {
 
     public static final String NCD_PROGRAM_CONCEPT_UUID = "099d3ba9-b302-4565-a060-09915f3c85b5";  // HUM NCD
     public static final String NCD_PROGRAM_OUTCOMES_CONCEPT_UUID = "a65eb5bf-d0a5-4ce3-b3c5-b16d0df3f468";  // 11977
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/pihcore/PihCoreConstants.java
```diff
@@ -97,7 +97,7 @@
 
     public static final String NCD_PROGRAM_CONCEPT_UUID = "099d3ba9-b302-4565-a060-09915f3c85b5";  // HUM NCD
     public static final String NCD_PROGRAM_OUTCOMES_CONCEPT_UUID = "a65eb5bf-d0a5-4ce3-b3c5-b16d0df3f468";  // 11977
-    public static final String NCD_PROGRAM_STATUS_UUID = "83a258b8-3294-4746-acb6-f045cccd37e8";
+    public static final String NCD_PROGRAM_STATUS_UUID = "83a258b8-3294-4746-acb6-f045cccd37e8";  // 11506
```
mogoodrich, line 100: Is 11506 a mapping code?  I would put "PIH:11506" or "CIEL:11506" to make it clear.

mogoodrich (changes requested):
Let's add "PIH:" before all the comments for concept codes to make it clear what they are referring to...

Otherwise looks good... I didn't check that all the workflow/states match what is in the tickets, but that will easy enough to do while testing...

brandones:
Fixed the comments. Merging. Thanks Mark!

PR #217, opened 2020-12-02
Split HIV and COVID components. Create TB component, along with Form and EncounterType

brandones:
https://pihemr.atlassian.net/browse/PER-351

Message code for new ET: https://github.com/PIH/openmrs-config-pihemr/commit/21784400d46ab5d65a96922c766b9971eba0986b

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
@@ -1,5 +1,8 @@
 package org.openmrs.module.pihcore.config;
 
+/**
+ * See https://github.com/PIH/openmrs-config-pihemr/#existing-components for documentation of these components
+ */
 public class Components {
 
         public static final String VISIT_MANAGEMENT = "visitManagement";
@@ -17,7 +20,7 @@ public class Components {
         public static final String SURGERY = "surgery";
         public static final String ONCOLOGY = "oncology";  // program
         public static final String NCD = "ncd";  // program
-        public static final String HIV = "hiv";  // program. Needs HIVProgramBundle and Haiti HIV MDS package
+        public static final String HIV = "hiv";
         public static final String HIV_PROGRAM = "hivProgram";
         public static final String HIV_FORMS = "hivForms";
         public static final String HIV_INTAKE_FORM = "hivIntakeForm";
--- a/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/config/Components.java
@@ -18,6 +18,9 @@ public class Components {
         public static final String ONCOLOGY = "oncology";  // program
         public static final String NCD = "ncd";  // program
         public static final String HIV = "hiv";  // program. Needs HIVProgramBundle and Haiti HIV MDS package
+        public static final String HIV_PROGRAM = "hivProgram";
+        public static final String HIV_FORMS = "hivForms";
+        public static final String HIV_INTAKE_FORM = "hivIntakeForm";
         public static final String VCT = "vct";
         public static final String MCH = "mch";  // mchForms + mchProgram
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -18,14 +18,19 @@
         public static final String ONCOLOGY = "oncology";  // program
         public static final String NCD = "ncd";  // program
         public static final String HIV = "hiv";  // program. Needs HIVProgramBundle and Haiti HIV MDS package
+        public static final String HIV_PROGRAM = "hivProgram";
+        public static final String HIV_FORMS = "hivForms";
+        public static final String HIV_INTAKE_FORM = "hivIntakeForm";
```
mogoodrich, line 23: Would be good to document here what each does:
* HIV_INTAKE_FORM: just Intake
* HIV_FORMS: all forms
* HIV_PROGRAM: just HIV Program
* HIV: HIV Program + All HIV Forms

... maybe even with the note that you should only ever have to include one of these per implementation (right?)  (It totally makes sense to me that HIV_FORMS should include the the intake form, so you don't have to enable both HIV_FORMS and HIV_INTAKE_FORMS, but I worry that without documentation people will set them all within a single implementation "just to safe" leading to confusing down the road).

api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -18,14 +18,19 @@
...
         public static final String VCT = "vct";
         public static final String MCH = "mch";  // mchForms + mchProgram
         public static final String MCH_FORMS = "mchForms";  // used by ZL (via "mch")
         public static final String MCH_PROGRAM = "mchProgram";  // needs MCHProgramBundle[ZL], used by ZL (via "mch")
         public static final String VACCINATION_FORM = "vaccination";
         public static final String ZIKA = "zika";  // program
         public static final String COVID19 = "covid19";
+        public static final String COVID19_INTAKE_FORM = "covid19IntakeForm";
```
mogoodrich, line 31: Same thing as above, would be good to note that should should only install COVID_INTAKE_FORM *or* COVID, depending on whether you want the full COVID set or just the intake form.

(And yes, it will be great when we have a easier/more flexible way to config on a form-by-form basis :) )

mogoodrich (changes requested):
I recommend adding some usage comments, but otherwise LGTM...

api/src/main/java/org/openmrs/module/pihcore/config/Components.java
brandones: Oh yeah, I'd meant to do that! Thanks.

brandones:
@mogoodrich https://github.com/PIH/openmrs-config-pihemr/commit/3d4eae8f12099488966acd4d84cf84492fa64eab

api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -18,14 +18,19 @@
         public static final String ONCOLOGY = "oncology";  // program
         public static final String NCD = "ncd";  // program
         public static final String HIV = "hiv";  // program. Needs HIVProgramBundle and Haiti HIV MDS package
```
mogoodrich, line 20: Let's remove the "program" part of this comment here, which was outdated before but now is particularly confusing with the "hivProgram" component as well below.

mogoodrich (approved):
LGTM with the updated documentation to config PIH-EMR... though I'd stil may the small tweak to the old comment for the "hiv" component (I added a comment).

aastupilloses:
I agree with @mogoodrich that those constants must be explained. Tbh, I'm not so much into this and I found very useful to know what they mean.

brandones:
Thanks @aastupilloses ! Note that all the components are documented in the [config-pihemr README](https://github.com/PIH/openmrs-config-pihemr/#existing-components).

PR #218, opened 2020-12-11
install all program bundles at all sites

mogoodrich:


mogoodrich:
@lnball  @brandones  thoughts on this?  This is just continuing to move toward a model where we have a larger set of core metadata installed in all PIH EMR implementations.  In this case, we are basically installing all programs in all implementations.

This came up after a discussion with Vamba where he requested just being able to turn a program on/off via the component (and not have to remember to install the Metadata as well), but it was also something I had been thinking about earlier.

Interested in your thoughts... if it's a go, I will probably wait until next merge to merge so I can test to make sure it works properly for each country.

lnball:
We still use components to have the programs "appear".  When @mogoodrich writes "install", it makes me think that they will all appear on every site and dashboard.

mogoodrich:
@lnball No, the component will still determine whether a program appears or not... the difference is the underlying program metadata will be installed behind-the-scenes everywhere.

mogoodrich:
(The problem previously was that, say you enable "xyzProgramComponent" but forgot to add the "xyzzProgramMetadataBundle" to your "CountryMetadataBundle" things would crash)

lnball:
Then it's a win for me.  Less chance to forget a step.

mogoodrich:
yeah, exactly @lnball

api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/mexico/MexicoMetadataToInstallAfterConceptsBundle.java
```diff
@@ -1,31 +0,0 @@
...
-// concepts have been installed via metadata sharing (since programs reference concepts)
-@Component
-@Requires({
-        ANCProgramBundle.class,
-        AsthmaProgramBundle.class,
-        DiabetesProgramBundle.class,
-        EpilepsyProgramBundle.class,
-        HypertensionProgramBundle.class,
-        MalnutritionProgramBundle.class,
-        MentalHealthProgramBundle.class
```
brandones, line 25: I think you need to add these to the PIH Programs bundle. As well as the Covid19 program.

brandones:
Yeah, definitely an improvement. Best would be moving all this to the Iniz programs loader. Then it's also easy for sites to override programs, e.g. to change the workflows or states.

mogoodrich:
I double checked and I believe they are all in the PIHProgramsBundle @brandones 

https://github.com/PIH/openmrs-module-pihcore/blob/master/api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/program/PihProgramsBundle.java

(Though I suspect they weren't at one point which is why you added them here)

And, yeah, Iniz is still the long-term plan.

brandones:
Aha! Great. Thanks.

brandones (approved):

mogoodrich:
Thanks @brandones  @lnball , merging this in.

Technically, we should try re-installing a SDK with each country from scratch, but I'm going to forgo that for now.

PR #219, opened 2020-12-21
Peru name template

MiguelAHPpih:
First time doing a pull request, sorry for taking so much time.
I added the "familyName2" in the first IF, i added it also on the second to force it but got no results.
PD: Forgot to delete the "familyName2" on the second IF

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/setup/NameTemplateSetup.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/setup/NameTemplateSetup.java
@@ -12,39 +12,74 @@ import java.util.Map;
 
 public class NameTemplateSetup {
 
-    public static void configureNameTemplate(NameSupport nameSupport, Config config) {
+    public static void configureNameTemplate(NameSupport nameSupport, Config config){
+        if(config !=null){
+            if (config.getCountry() == ConfigDescriptor.Country.PERU) {
+                NameTemplate nameTemplate = new NameTemplate();
+                nameTemplate.setCodeName("short");  // we are redefining the short name template for use in our context
 
-        if (config.getCountry() == ConfigDescriptor.Country.PERU) {
-        } else {
+                Map<String, String> nameMappings = new HashMap<String, String>();
+                nameMappings.put("givenName", "zl.givenName");
+                nameMappings.put("familyName", "zl.familyName");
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/pihcore/setup/NameTemplateSetup.java
```diff
@@ -10,37 +12,75 @@
...
-        // we blow away the other templates here, is that a bad thing?
-        nameSupport.setLayoutTemplates(templates);
-        nameSupport.setDefaultLayoutFormat("short");
+                NameTemplate nameTemplate = new NameTemplate();
+                nameTemplate.setCodeName("short");  // we are redefining the short name template for use in our context
 
+                Map<String, String> nameMappings = new HashMap<String, String>();
+                nameMappings.put("givenName", "zl.givenName");
+                nameMappings.put("familyName", "zl.familyName");
+                nameMappings.put("familyName2", "zl.familyName2");
```
brandones, line 57: Please leave the non-Peru name template as it was. So no `familyName2`.

api/src/main/java/org/openmrs/module/pihcore/setup/NameTemplateSetup.java
```diff
@@ -10,37 +12,75 @@
...
+                NameTemplate nameTemplate = new NameTemplate();
+                nameTemplate.setCodeName("short");  // we are redefining the short name template for use in our context
 
-        NameTemplate nameTemplate = new NameTemplate();
-        nameTemplate.setCodeName("short");  // we are redefining the short name template for use in our context
+                Map<String, String> nameMappings = new HashMap<String, String>();
+                nameMappings.put("givenName", "zl.givenName");
+                nameMappings.put("familyName", "zl.familyName");
+                nameMappings.put("familyName2", "zl.familyName2");
+                nameMappings.put("middleName", "zl.nickname");
```
brandones, line 25: Didn't you want to get rid of the middle/nick name box?

mogoodrich (approved):
LGTM, with the same comments that @brandones  made but it looks like you just resolved them in a follow-on commit.

I'm going to go ahead and merge so we can experiment/test out as well... it may just be, unfortunately, that the Registration App doesn't yet support the familyName2 field.

PR #226, opened 2021-01-28
UHM-5112: For Intake and Followup Forms, migrate Form Version

mogoodrich:


omod/src/main/webapp/pages/visit/templates/encounters/hivNotMigratedEncounterLong.gsp
```diff
@@ -0,0 +1,30 @@
+<div class="header" ng-include="'templates/encounters/defaultEncounterHeader.page'">
+</div>
+
+<div class="content encounter-summary-long">  <!-- encounter-summary-long currently only used for Selenium tests -->
+    <p class="bold">
+        ${ ui.message("pihcore.hiv.legacyDataNotMigrated") }
+    </p>
```
mogoodrich, line 7: This is basically just a copy of defaultEncounterLong template with the above message added... we can tweak as needed as others review.

omod/src/main/webapp/resources/scripts/visit/filters.js
```diff
@@ -398,7 +398,7 @@ angular.module("filters", [ "uicommons.filters", "constants", "encounterTypeConf
 
             if (encounters) {
                 _.each(encounters, function(e) {
-                    if (EncounterTypeConfig.get(e.encounterType.uuid) && EncounterTypeConfig.get(e.encounterType.uuid).showOnVisitList
+                    if (EncounterTypeConfig.get(e) && EncounterTypeConfig.get(e).showOnVisitList
```
mogoodrich, line 401: changing the first parameter of the encounterTypeConfig.get so that it takes the full encounter, not the encounter type.

(Refactored get method coming in the config-pihemr PR)

omod/src/main/webapp/resources/scripts/visit/visit.js
```diff
@@ -491,6 +491,8 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                  ngDialog, Encounter, EncounterTypeConfig, AppFrameworkService, visitUuid, visitTypeUuid, suppressActions, patientUuid, encounterUuid,
                  locale, currentSection, country, site, DatetimeFormats, EncounterTransaction, SessionInfo, Concepts, VisitTypes) {
 
+          const visitRef = "custom:(uuid,startDatetime,stopDatetime,location:ref,encounters:(uuid,display,encounterDatetime,patient:default,location:ref,form:(uuid,version),encounterType:ref,obs:default,orders:ref,voided,visit:ref,encounterProviders:(uuid,encounterRole,provider,dateCreated),creator:ref),patient:default,visitType:ref,attributes:default)"
+
```
mogoodrich, line 495: Changing the visit ref to include the form version

brandones (approved):

omod/src/main/webapp/pages/visit/templates/encounters/hivNotMigratedEncounterLong.gsp
```diff
@@ -0,0 +1,30 @@
...
+    <p ng-repeat="obs in encounter.obs" class="aligned">
+        <label>{{ obs.concept | omrsDisplay }}</label>
+        <span class="value">{{ obs | obs:"value" }}</span>
+    </p>
+    <div ng-show="encounter.orders && encounter.orders.length">
+        <span ng-repeat="order in encounter.orders">
+            <p class="aligned">
+                <label>${ ui.message("coreapps.patientDashBoard.orderNumber")}</label>
+                <span class="value">{{ order.orderNumber }}</span>
```
mseaton, line 16: As far as our migrated data is concerned, no order numbers are meaningful.  For drug orders (which are the only orders we are migrating into afaik), this is a generated number that we are inventing just so this is non-null.  So I'd prefer we not include it.  I also don't think urgency is particularly useful (below).  Just displaying "effectiveStartDate" will give you either the date already started or the date scheduled to start (whether that is in the past or future) which is typically what we want to see I think.   I don't know what "order.display" is - I'm not aware of any "display" method or property on Order.  I'd think here we'd want to show the orderable (concept).

omod/src/main/webapp/pages/visit/templates/encounters/hivNotMigratedEncounterLong.gsp
```diff
@@ -0,0 +1,30 @@
+<div class="header" ng-include="'templates/encounters/defaultEncounterHeader.page'">
+</div>
+
+<div class="content encounter-summary-long">  <!-- encounter-summary-long currently only used for Selenium tests -->
+    <p class="bold">
+        ${ ui.message("pihcore.hiv.legacyDataNotMigrated") }
+    </p>
+    <p ng-repeat="obs in encounter.obs" class="aligned">
+        <label>{{ obs.concept | omrsDisplay }}</label>
+        <span class="value">{{ obs | obs:"value" }}</span>
```
mseaton, line 10: I don't know if this has any special functionality that isn't obvious here, but I'd think we'd want to try, at a minimum, to at least organize Obs by Obs Group if possible (display them together in some sort of field set maybe, with the grouping obs as the field set header).

mseaton (approved):
LGTM, just some comments on how we should improve our default display of encounters, particularly around migrated orders.

mogoodrich:
I just copied over the "defaultEncounterLong.gsp" template that we created years ago for Mirebalais and added the warning message... basically just to use as a starting point as to what our "view" mode for these encounters should display.

I agree though that the order stuff doesn't make much/any sense in this context, as the migrated Intake and Followup encounter types should not have an orders associated with them.  I'll pull that.

I agree that we probably can improve the bare-bones obs display, but I'm going to leave that as-is for now, and we can have further discussions... don't want to put effort into it, if, for example, we do just decide to build out the old forms.

mogoodrich:
Simplified the notMigrate template to just show obs for now... we can review and tweak as necessary...

PR #231, opened 2021-02-13
UHM-4939 - Add ability to specify the column configuration of the fin…

mseaton:
…d patient widget through pih config.


mogoodrich (approved):
LGTM!

PR #233, opened 2021-02-23
Sl 185

mseaton:
Adds initial pass at KGH location and primary identifier with identifier source, and adds a new identifier generator provider for this KGH Identifier source.  Requires latest idgen snapshot.


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #234, opened 2021-02-24
Modify existing code for dev id setup to reflect recommended approach…

mseaton:
… on readme

I think we had all forgotten or never realized that there is code in place already to enable a different set of identifier sources to be configured in development and test environments.  I wasn't able to exactly get this working as written when I tried it out - it resulted in errors when trying to connect from localhost to localhost for some reason.  But I think the principle is a good idea that we want to provide an option to get set up with the right identifier source configuration based on a runtime property rather than the documented process one needs to currently take in the wiki.  These small tweaks take the current (not working) code and tweaks it slightly such that it should set up the recommended dev/test setup.

Basically, if you add:

local_zl_identifier_generator_enabled=true
local_zl_identifier_generator_prefix=Y

to your openmrs-runtime.properties file before starting up the full distribution, then it will create a local identifier generator and connect it up to the pool, instead of creating a remote source and connecting it up to the pool.


cioan (approved):

mogoodrich (approved):
LGTM, thanks @mseaton !

PR #240, opened 2021-03-24
SL-191 - Add "KGH Triage" location

mseaton:


api/src/main/java/org/openmrs/module/pihcore/metadata/sierraLeone/SierraLeoneLocations.java
```diff
@@ -33,4 +33,10 @@ public LocationDescriptor parent() {
         public String name() { return "KGH"; }
         public String description() { return "KGH"; }
     };
+
+    public static LocationDescriptor KGH_TRIAGE = new LocationDescriptor() {
+        public String uuid() { return "2bcb9215-8cd6-11eb-b7be-0242ac110002"; }
+        public String name() { return "KGH Triage"; }
+        public String description() { return "KGH Triage"; }
```
mogoodrich, line 40: Should be a parent of KGH...

mogoodrich (changes requested):

api/src/main/java/org/openmrs/module/pihcore/metadata/sierraLeone/SierraLeoneLocations.java
mseaton: Yeah, I noticed that too.  Already committed.  Tahnks

cioan (approved):

PR #242, opened 2021-03-31
UHM-5160 - Add unit test that enables running all of the OpenMRS vali…

mseaton:
…dators against a given database.


mseaton:
@mogoodrich - I went ahead and merged this, as it's just a single new unit test. Review still welcome, but I want to get this into some test usage and that is easier than dealing with a feature branch.

PR #243, opened 2021-04-01
UHM-4169 - Update code the loads in global scripts and styles to ensu…

mseaton:
…re file paths used are compatible with the ConfigurationResourceProvider using relative paths


lnball (commented):
How can you test this @mseaton ?  Thanks for diving into this.  Much appreciated.

mseaton:
I've been trying to figure that out @lnball .  @mogoodrich said he'd do it, but easiest to me would be to just replace the pihcore module on humci or ci with one built from this branch and see what happens, if we can spare either for a bit.

mogoodrich:
We can just merge this in once I have a chance to review.  Patience please.

mogoodrich (approved):
LGTM... since there seems to a a rush to get this in, will merge (and then revert if it breaks things)

PR #247, opened 2021-04-13
UHM-4354 (Pt 1, just Mexico) Move Location, Location Tag, and Lcation Attribute Types installation out of PIH core and into config repos

brandones:
A first step toward https://pihemr.atlassian.net/browse/UHM-4354 . This moves Location Tags and Attribute Types out to configuration. It completely moves Mexico locations out to configuration.

I tested local servers with both CES and ZL HIV configurations.

Depends on https://github.com/PIH/openmrs-config-pihemr/pull/162 and https://github.com/PIH/openmrs-config-ces/pull/24 and https://github.com/PIH/openmrs-module-mirebalais/pull/147


api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/LocationBundle.java
```diff
@@ -6,7 +6,7 @@
 import org.springframework.stereotype.Component;
 
 @Component
-@Requires({ LocationTagBundle.class, LocationAttributeTypeBundle.class} )
+@Requires({ LocationAttributeTypeBundle.class} )
```
mogoodrich, line 9: I saw you added the attribute types to Iniz in config-pihemr... is there a reason you are still installing them here?

mogoodrich (commented):
One question, but generally looks good.

I like the approach of moving the Location Tags and Attributes out and starting with a single country (Mexico), confirming that works, and then moving on to other countries from there.

We also should test a Mexico install from "scratch" (though that could happen after merging)

api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/LocationBundle.java
brandones: Oversight. Thanks!

mseaton (approved):
Nice!  LGTM.

PR #250, opened 2021-04-21
UHM-4354 - Remove Mexico and Sierra Leone locations, moving to config

brandones:
https://pihemr.atlassian.net/browse/UHM-4354

Depends on https://github.com/PIH/openmrs-config-pihsl/pull/11


api/src/main/java/org/openmrs/module/pihcore/identifier/sierraLeone/ConfigureSierraLeoneIdGenerators.java
```diff
@@ -31,7 +29,7 @@ public static void configurePrimaryIdentifierSource(IdentifierSourceService iss,
                     9,
                     "1234567890",
                     "100001",
-                    SierraLeoneLocations.WELLBODY_CLINIC
```
mogoodrich, line 34: I understand the desire not to create a constant for a single use, but then maybe a comment here saying something like "uuid for Wellbody Clinic, defined in locations.csv in config-pihsl"?

api/src/main/java/org/openmrs/module/pihcore/identifier/sierraLeone/ConfigureSierraLeoneIdGenerators.java
```diff
@@ -51,7 +49,7 @@ else if ("KGH".equalsIgnoreCase(config.getSite())) {
                     12,
                     "0123456789",
                     "0001",
-                    SierraLeoneLocations.KGH
```
mogoodrich, line 54: and same here...

mogoodrich (approved):
A request to add a couple comments, but otherwise LGTM

PR #251, opened 2021-04-21
UHM-4425, install app privileges via iniz config

cioan:


mogoodrich:
Or, actually, looking at the other PR, maybe you are leaving the Roles for now and just testing moving the Privileges?  This makes sense to me!

PR #254, opened 2021-04-26
UHM-4425, install Liberia roles via the initializer

cioan:


brandones (approved):
Love to see it. Would you mind linking these to the corresponding config changes, for easy comparison?

brandones:
Found it

PR #258, opened 2021-05-05
UHM-5486: Invalid UUID associated with program workflow state

mseaton:


mogoodrich (commented):
This looks good to me, but I think we should wait to merge this in until we merge in the full program changes, so we can test all these changes going out a server in a single release, as will likely happen in a production environment? (ie, this will help us confirm that the uuid change happens before the Iniz install)

PR #261, opened 2021-05-13
UHM-4350: Remove custom Mexico metadata bundles to config-ces

mseaton:


api/src/main/java/org/openmrs/module/pihcore/identifier/mexico/ConfigureMexicoIdGenerators.java
```diff
@@ -38,7 +38,7 @@ private static void createPrimaryIdentifierGenerator(IdentifierSourceService ide
         chiapasPrimaryIdentifierSource.setName("Chiapas Primary Identifier Source");
         chiapasPrimaryIdentifierSource.setDescription("Primary Identifier Generator for Chiapas");
         chiapasPrimaryIdentifierSource.setIdentifierType(MetadataUtils.existing(PatientIdentifierType.class,
-                MexicoPatientIdentifierTypes.CHIAPAS_EMR_ID.uuid()));
+                CesConfigConstants.PATIENTIDENTIFIERTYPE_CHIAPASEMRID_UUID));
```
mogoodrich, line 41: Nit, but I'm wondering if it would make sense to have an underscore in the constants based on both camel case and the level in the yaml file?  (If I'm understanding it, it's only the level in the yaml file now?)

Probably fine how it is, just throwing it out there.

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataMappingsSetup.java
```diff
@@ -80,6 +80,11 @@ public static void setupPrimaryIdentifierType(MetadataMappingService metadataMap
         metadataMappingService.mapMetadataItem(patientIdentifierType, EmrApiConstants.EMR_CONCEPT_SOURCE_NAME, EmrApiConstants.PRIMARY_IDENTIFIER_TYPE);
     }
 
+    public static void setupPrimaryIdentifierType(MetadataMappingService metadataMappingService, PatientService patientService, String identifierTypeUuid) {
+        PatientIdentifierType patientIdentifierType = patientService.getPatientIdentifierTypeByUuid(identifierTypeUuid);
+        metadataMappingService.mapMetadataItem(patientIdentifierType, EmrApiConstants.EMR_CONCEPT_SOURCE_NAME, EmrApiConstants.PRIMARY_IDENTIFIER_TYPE);
+    }
```
mogoodrich, line 86: Maybe I should have looked at the CES config PR closer, but it this removed because Iniz supports setting up metadata mappings (great, if so!)

mogoodrich (approved):
A couple comments, but looks good!

api/src/main/java/org/openmrs/module/pihcore/identifier/mexico/ConfigureMexicoIdGenerators.java
mseaton: Yeah.  We'd have to update the plugin and then update all of the constant references.  If it's important to us, we can do so, but let's punt.

api/src/main/java/org/openmrs/module/pihcore/setup/MetadataMappingsSetup.java
mseaton: Nothing is removed.  I just added a new method that takes the uuid rather than the descriptor, as the descriptor went away and the uuid is available.

mogoodrich: Ah, got it, missed that...

api/src/main/java/org/openmrs/module/pihcore/identifier/mexico/ConfigureMexicoIdGenerators.java
mogoodrich: Agreed, let's punt!

PR #266, opened 2021-05-18
Use npx sass for sass compilation, not ruby

brandones:
I couldn't get live reloading to work. When I run `mvn package` it does recompile the CSS, but the changes don't appear in the server on refreshing the page. Would love it if someone could carry this across the finish line.


mogoodrich (approved):
LGTM, from my limited knowledge.  Will be great to get rid of the Ruby dependency... looks like @mseaton  has picked it up for taking it over the finish line?

mseaton (approved):
Great to get rid of Ruby, I don't know enough about the changes to be able to review sufficiently.  I think it'll need to be tested.  @mogoodrich I didn't pick this up - I think Brandon assigned it to me, so feel free to take it over from me.

PR #275, opened 2021-05-26
UHM-5554:  Migrate order frequencies to Initializer

mseaton:


api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/OrderEntryConcepts.java
```diff
@@ -1,103 +0,0 @@
...
-        install(orderFrequency(Concepts.WHEN_REQUIRED, null, "da558e86-3fbc-427a-8e4e-06d61c38d813"));
-        install(orderFrequency(Concepts.IMMEDIATELY, null, "8487d0db-3378-49b8-b24c-18f121dfbfd0"));
-    }
-
-    /**
-     * OrderFrequency is not allowed to be resaved after it has an order pointing to it, or it will throw the exception:
-     * "org.openmrs.api.APIException: This order frequency cannot be edited because it is already in use"
-     *
-     * This method only calls the base install method if a field has actually changed, to prevent against spuriously
-     * getting this error.
```
mogoodrich, line 88: @mseaton see comment above... will Iniz be smart enough to not run into this?  (I guess you probably already tested)

omod/src/test/java/org/openmrs/module/pihcore/htmlformentry/HtmlFormComponentTest.java
```diff
@@ -45,9 +44,6 @@ public void beforeEachTest() throws Exception {
         executeDataSet("requiredDataTestDataset.xml");
         authenticate();
 
-        // load the API test bundle of MDS concepts
-        metadataDeployService.installBundle(Context.getRegisteredComponents(OrderEntryConceptsFromMetadataSharing.class).get(0));
```
mogoodrich, line 49: Do we just not need to load this for the tests, or am I missing the place where we start loading them via config instead?

mogoodrich (approved):
A couple comments, but LGTM... I assume the change where PIH Core calls Iniz explicity to do the loading is going to show up the PR for the other ticket?

api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/OrderEntryConcepts.java
mseaton: I don't have any idea actually

omod/src/test/java/org/openmrs/module/pihcore/htmlformentry/HtmlFormComponentTest.java
mseaton: The tests still passed after removing, so...

api/src/main/java/org/openmrs/module/pihcore/deploy/bundle/core/OrderEntryConcepts.java
mogoodrich: We will have to keep an eye on this, then, I guess... if your dev environments didn't have any orders in them you may not have run into this?

mseaton: Good catch.  I created an issue in initailizer github, and a PR/branch that demonstrates it, and created an issue in our JIRA to keep track of this.

PR #276, opened 2021-05-27
UHM-5558:  Migrate global metadata mappings to initializer

mseaton:


api/src/test/java/org/openmrs/module/pihcore/PihCoreContextSensitiveTest.java
```diff
@@ -4,22 +4,42 @@
...
+    public PihCoreContextSensitiveTest() {
+        super();
+        {
+            Module mod = new Module("", "metadatamapping", "", "", "", "1.3.4");
+            mod.setFile(new File(""));
+            ModuleFactory.getStartedModulesMap().put(mod.getModuleId(), mod);
+        }
+    }
+
+    @Override
```
mseaton, line 37: I spent a really long time fighting unit tests today, and this is what I was able to come up with to get everything running and passing, unfortunately it really slows things down.  We can investigate and try to improve later on, but for now this works.  Basically I think we need to add the metadatamapping module like this to get Initializer to load in metadata mappings in the unit tests, but then doing so somehow messes up tests when run in the suite.  I was able to mostly fix that with the DirtiesContext annotation, but then a few tests started randomly failing and prompting user/password login prompts.  Adding the deleteAllData for all tests as below ended up fixing this, but this is pretty brute force.

mogoodrich (approved):
LGTM... would be good to have @brandones  and @cioan  take a look, especially as an fyi about the change to the way we are loading from Iniz (ie, doing it all manually from here)

PR #278, opened 2021-05-27
UHM-5533, encode file path

pihinformatics:
@mogoodrich, this seems to take care of the Windows path issue.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
@@ -42,9 +42,9 @@ public class PihCoreUtil {
 
     public static String getFormResource(String formName) {
         try {
-            return URLEncoder.encode("file:" + getFormDirectory() + formName, StandardCharsets.UTF_8.toString()).toString();
+            return URLEncoder.encode("file:" + getFormDirectory() + formName, StandardCharsets.UTF_8.toString());
         } catch (UnsupportedEncodingException e) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
```diff
@@ -37,7 +41,11 @@ public static String getDrugListChecksumDirectory() {
     }
 
     public static String getFormResource(String formName) {
-        return "file:" + getFormDirectory() + formName;
+        try {
+            return URLEncoder.encode("file:" + getFormDirectory() + formName, StandardCharsets.UTF_8.toString()).toString();
```
mogoodrich, line 45: Doesn't URLEncoder.encode return a string (ie, why do we need the toString at the end?)

api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
```diff
@@ -37,7 +41,11 @@ public static String getDrugListChecksumDirectory() {
...
+        } catch (UnsupportedEncodingException e) {
+            return "file:" + getFormDirectory() + formName;
```
mogoodrich, line 47: Is there a use case here where this wouldn't be an error?  ie, would it be better just to throw an exception here?

mogoodrich (commented):
Thanks @cioan  for tracking this down!  I added a couple questions.

mogoodrich:
I went ahead and made these tweaks, going to merge this in.

PR #281, opened 2021-06-02
UHM-5548:  Migrate concept sources to Initializer

mseaton:


mogoodrich (approved):
Didn't directly compare between the new and old,  but generally looks good to me!

PR #282, opened 2021-06-30
UHM-5624 Removed 3 concept.java files -- all the concepts are in mds …

lnball:
Please review this code and let me know if I missed anything. @mogoodrich 

I tested with my local SDK -- with a database and with a clean database.  Not sure how to confirm that it's using the latest pihcore module. I set a watch on the module as soon as I started.


mogoodrich (approved):
LGTM, thanks for doing this @lnball

PR #284, opened 2021-07-26
UHM-5627:Do not automatically close visit if it was created within th…

mogoodrich:
…e past 24 hours


brandones:
The code looks good. The logic for deciding when to close a visit is a little confusing (what's `adtService.shouldBeClosed`? Isn't that entire 4-line if statement trying to evaluate whether the visit should be closed?). But that's just to say that it could maybe use a refactor at some point, or a comment.

LGTM.

brandones (approved):

mogoodrich:
Yep, it's a mess... as a history, adtService.shouldBeClosed is a method that is provided by EmrApiService to determine whether the visit should be closed, which was all we used at first.  (In fact, there's a regular CloseStaleVisitTask I believe that still uses it).

And then the ED department at Mirebalais requested all this more complex logic.  

Anyway, thanks!

PR #286, opened 2021-08-12
Don't use LATEST version for config repos

brandones:
Use the explicit snapshot versions. These are kept up to date by the release job: https://github.com/PIH/openmrs-config-pihemr/pull/235 . The latter commit on this PR was made by the release job.


mogoodrich (approved):
LGTM!

PR #288, opened 2021-08-17
UHM-5721: Bug with logic to keep visit open if created or changed in …

mogoodrich:
…the past 24 hours

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
@@ -1,6 +1,8 @@
 package org.openmrs.module.pihcore.task;
 
 import org.apache.commons.lang.time.DateUtils;
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
 import org.openmrs.Encounter;
 import org.openmrs.Location;
 import org.openmrs.LocationTag;
@@ -15,7 +17,6 @@ import org.openmrs.module.emrapi.disposition.DispositionType;
 import org.openmrs.module.emrapi.visit.VisitDomainWrapper;
 import org.openmrs.module.pihcore.PihEmrConfigConstants;
 import org.openmrs.scheduler.tasks.AbstractTask;
-import org.openmrs.util.OpenmrsUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


api/src/test/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTaskTest.java
```diff
@@ -519,7 +519,55 @@ public void test_shouldNotKeepVisitOpenIfItHasEDTriageAndEncounterWithinFortyEig
...
+        visit.setDateChanged(new Date());
+        visit.setStartDatetime(DateUtils.addHours(new Date(), -60));
+        visit.setPatient(patient);
+        visit.setLocation(location);
+        visit.setVisitType(emrApiProperties.getAtFacilityVisitType());
+
+        // create an exit from care encounter
+        Encounter exitFromCareEncounter = new Encounter();
+        exitFromCareEncounter.setPatient(patient);
+        exitFromCareEncounter.setEncounterType(encounterService.getEncounterTypeByUuid("b6631959-2105-49dd-b154-e1249e0fbcd7"));
```
brandones, line 550: You couldn't get it by name?

api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
```diff
@@ -54,13 +55,65 @@ public void execute() {
...
+            VisitDomainWrapper wrappedVisit = adtService.wrap(visit);
+            Boolean changedOrUpdatedRecently = changedOrUpdatedRecently(visit, REGULAR_VISIT_EXPIRE_TIME_IN_HOURS);
+
+            Disposition mostRecentDisposition = wrappedVisit.getMostRecentDisposition();
+            Long hoursSinceLastEncounter = wrappedVisit.getMostRecentEncounter() != null ?
+                    new Duration(new DateTime(wrappedVisit.getMostRecentEncounter().getEncounterDatetime()), new DateTime()).getStandardHours()
+                    : null;
+
+            // **this logic is intentionally verbose, since it can be difficult to follow**
+            Boolean closeVisit = false;
```
brandones, line 68: Nice way to approach this. I can't fit the whole thing in my brain but I trust each line, more or less. If you're uneasy at all about the mutable state variable `closeVisit` you could get rid of it by factoring this whole thing out into a function and making each of those `closeVisit = x` lines into `return x`.

brandones (approved):

api/src/test/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTaskTest.java
mogoodrich: Even better, turns out I can just get it by EmrApiProperties, I just updated...

PR #292, opened 2021-08-24
UHM-5628: Properly handle visit when admission entered retrospectively

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
```diff
@@ -40,4 +45,27 @@ public static String getFormResource(String formName) {
         return "file:" + getFormDirectory() + formName;
     }
 
+    /**
+     * Utility  method that reopens a visit (sets its end date to null) if it is the most recent visit for a patient
+     * (Used by the ReopenVisitAction and ReopenVisitDispositionAction)
+     * TODO: is there a better place for this to live?
```
brandones, line 51: I would put it in the directory with the action classes

brandones (approved):
LGTM

api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
mogoodrich: Generally agree, but the action classes were already grouped into different directories... but, as a I idea, maybe I can refactor that part of it

PR #294, opened 2021-08-31
PETL-90:  Store enabled components in the database to enable PETL to …

mseaton:
…use components within conditionals to determine whether or not to execute a job.


mogoodrich (approved):
LGTM!

PR #298, opened 2021-09-03
UHM-5709:  Support require expressions that limit to age in months

mseaton:


omod/src/main/webapp/resources/scripts/visit/filters.js
```diff
@@ -483,10 +486,16 @@ angular.module("filters", [ "uicommons.filters", "constants", "encounterTypeConf
               return list.some(func);
             };
 
-            // TODO we really should have a better place for utility functions like this
```
mogoodrich, line 486: Why did we get rid of this function?

Agree that it's confusing how I did it, but we will need a replacement for this?

mogoodrich: Ah, never mind, I see your changes to pihemr now.

PR #299, opened 2021-09-08
UHM-5724:  Allow initializer config files to be conditionally loaded …

mseaton:
…based on the pih site that is set.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
@@ -97,7 +97,7 @@ public class PihCoreActivator extends BaseModuleActivator implements DaemonToken
                 config = Context.getRegisteredComponents(Config.class).get(0); // currently only one of these
             }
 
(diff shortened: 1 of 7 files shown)


mogoodrich (commented):
Actually, so the problem with this is we need to make sure that "locations.csv" loads before "locations_mirebalais.csv", for example, and I don't think this code does that?

We are going to need to define the top-level locations like "Mirebalais" and "Hinche" in both the site-specific files and the main location file... those I'm realizing as I write this that there's a flaw in that design, because if there's a change to the locations.csv but not locations_mirebalais.csv the location tags for Mirebalais would get (incorrectly) reverted.

mseaton:
> Actually, so the problem with this is we need to make sure that "locations.csv" loads before "locations_mirebalais.csv", for example, and I don't think this code does that?

This should be handled by initializer.  The csv files in the locations domain are loaded in the order specified by a special "_order:X" control header.  See how I set these in the config-zl project.

mogoodrich:
Oh, awesome!   I am still realizing the flaw in my design though (that seems like it would occur frequently) ... any thoughts on that?

PR #303, opened 2021-09-14
UHM-5404 - Adds 2 features to support this ticket.

mseaton:
UHM-5404 - Adds 2 features to support this ticket.  1. Adds a new page controller that supports passing a patient identifier and type, and a particular set of configurations, to open a patient dashboard for a patient with that identifier.  2.  Adds the ability to add arbitrary extensions via PIH config as an alternative to doing this programmatically in the CALF.

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/page/controller/PatientPageController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/page/controller/PatientPageController.java
@@ -1,0 +1,62 @@
+package org.openmrs.module.pihcore.page.controller;
+
+import org.openmrs.Patient;
+import org.openmrs.PatientIdentifier;
+import org.openmrs.PatientIdentifierType;
+import org.openmrs.api.PatientService;
+import org.openmrs.ui.framework.UiUtils;
+import org.openmrs.ui.framework.annotation.SpringBean;
+import org.openmrs.ui.framework.converter.util.ConversionUtil;
+import org.openmrs.ui.framework.page.PageModel;
+import org.openmrs.ui.framework.page.Redirect;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import java.io.IOException;
+import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


omod/src/main/java/org/openmrs/module/pihcore/page/controller/PatientPageController.java
```diff
@@ -0,0 +1,62 @@
...
+import org.openmrs.ui.framework.page.Redirect;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class PatientPageController {
```
mogoodrich, line 20: Might be good to add a comment as to where/why we are using this, as years down the road we will probably forget.

Also, I'm probably missing something obvious, but I'm not 100% why this is needed at all? Is the full patient object not available in the Core Apps context where you render the link (that is, could the link just have a reference to the patient uuid)?

mogoodrich (approved):
One question about the need for a specific patient page to load patient by identifier, but generally looks good to me... thanks for adding that ability to set up extensions via the config, that's a pattern we might get some legs from.

omod/src/main/java/org/openmrs/module/pihcore/page/controller/PatientPageController.java
mseaton: Thanks @mogoodrich - comment added.  Per the comment, the reason this is needed is because we are not linking to a page in the same EMR instance, so the uuid and primary key id of the patient you need to link to is unknown.  The shared information that links the patient across the 2 systems is the HIVEMR-v1 identifier that is present in both.

mogoodrich: Duh, right, thanks @mseaton !

PR #304, opened 2021-09-15
UHM-5724

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
```diff
@@ -51,6 +55,27 @@ public static void loadPreConceptDomains() {
...
+     * Any config files that contain a "-site-", and which do not end with the site name, are excluded
+     */
+    public static List<String> getExclusionsForLoader(Loader loader, Config config) {
+        List<String> exclusions = new ArrayList<>();
+        if (loader instanceof BaseFileLoader) {
+            BaseFileLoader ll = (BaseFileLoader) loader;
+            String site = config.getSite() == null ? "" : config.getSite().toLowerCase();
+            for (File f : ll.getDirUtil().getFiles("csv")) {
+                String filename = f.getName().toLowerCase();
+                if (filename.contains("-site-") && !filename.endsWith("-" + site + ".csv")) {
```
mogoodrich, line 70: @mseaton here's the change I made...

mogoodrich:
@mseaton I'm reopening this as I plan to use it with the new locationtagmaps domain... I changed it so that site-specific files need to have "-site-" in the name (not just a hypen).

api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
mseaton: Seems ok.  Why not require file to end with "-site-" + site + ".csv"?

api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
```diff
@@ -42,7 +45,8 @@ public static void loadPreConceptDomains() {
             for (Loader loader : Context.getService(InitializerService.class).getLoaders()) {
                 if (!excludeList.contains(loader.getDomainName())) {
                     log.warn("Loading from Initializer: " + loader.getDomainName());
-                    loader.loadUnsafe(new ArrayList<>(), true);
+                    List<String> exclusionsForLoader = getExclusionsForLoader(loader, config);
+                    loader.loadUnsafe(exclusionsForLoader, true);
```
mseaton, line 49: Since you are re-opening this, we should really add this same logic here to the "loadPostConceptDomains" method, so that we have full coverage with consistent behavior.  Would you mind adding that in?

mogoodrich: Makes sense, will do.

api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
mogoodrich: Makes sense, will change.

api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
mseaton: It doesn't look like you did this?  I don't see anything below in the PR @mogoodrich

mseaton (changes requested):
See my comment...

api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
mogoodrich: Right, I haven't done it yet... :)

Or, rather, I did something, but haven't pushed it up yet...

mogoodrich:
@mseaton  I pushed up my latest changes... note that I didn't do any testing (or review, really) of the addition to the post concept domains... can you take a look?  Thanks!

mseaton (approved):

mogoodrich:
Merging this in, thanks @mseaton !

PR #307, opened 2021-09-17
UHM-5770: Refactor location tag configuration out of Iniz setup and i…

mogoodrich:
…nto PIH Config

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
@@ -95,8 +95,8 @@ public class PihCoreActivator extends BaseModuleActivator implements DaemonToken
                 config = Context.getRegisteredComponents(Config.class).get(0); // currently only one of these
             }
 
+			LiquibaseSetup.setup();
             InitializerSetup.loadPreConceptDomains(config);
-
             setDispositionConfig(config);
             setGlobalProperties(config);
             setExtraIdentifierTypes(metadataMappingService, patientService, config);
@@ -106,7 +106,6 @@ public class PihCoreActivator extends BaseModuleActivator implements DaemonToken
             MetadataMappingsSetup.setupPrimaryIdentifierTypeBasedOnCountry(metadataMappingService, patientService, config);
             PatientIdentifierSetup.setupIdentifierGeneratorsIfNecessary(identifierSourceService, locationService, config);
             PacIntegrationSetup.setup(config);
-            LiquibaseSetup.setup();
             AttachmentsSetup.migrateAttachmentsConceptsIfNecessary(conceptService);
            // RetireProvidersSetup.setupRetireProvidersTask();
             GlobalResourceSetup.includeGlobalResources();
--- a/api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
@@ -1,9 +1,12 @@
 package org.openmrs.module.pihcore.setup;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openmrs.module.pihcore.PihCoreUtil;
 import org.openmrs.util.DatabaseUpdater;
+
+import java.io.File;
 
 public class LiquibaseSetup {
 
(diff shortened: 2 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -95,8 +95,8 @@ public void started() {
                 config = Context.getRegisteredComponents(Config.class).get(0); // currently only one of these
             }
 
+			LiquibaseSetup.setup();
```
mogoodrich, line 98: I moved this up to happen before the Iniz setup, as I'll need it to clear out the location tag maps before the Iniz setup runs.

(Tangent: it occurs to me that it would be easy and helpful to add support for liquibase changesets that run both before and after metadata is loading... however, note that even before my change the liquibase changesets run before concept loading... )

api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
```diff
@@ -1,17 +1,22 @@
...
 public class LiquibaseSetup {
 
     protected static Log log = LogFactory.getLog(LiquibaseSetup.class);
 
     public static void setup() {
         try {
-            DatabaseUpdater.executeChangelog(PihCoreUtil.getLiquibaseChangeLog(), null);
+			for (File file : FileUtils.listFiles(new File(PihCoreUtil.getLiquibaseDir()), new String[]{"xml"}, true)) {
+				DatabaseUpdater.executeChangelog(file.getAbsolutePath(), null);
+			}
```
mogoodrich, line 19: Changed this to run all files in the "liquibase" directory instead of just "liquibase.xml".  This allowed me to define a custom "liquibase-haiti.xml" file that I put in the config-zl project.

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mseaton: Sounds fine, but I just assumed you would add columns for every location tag to the csv, with null/empty values, and it would take care of doing all of the necessary cleanup since all columns would be present.

api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
mseaton: Hmm. I'm not sure about this.  With a single liquibase file, one can control the order in which things execute.  And if you need multiple files, you can add a changeset to a file that says to execute another file.  With multiple arbitrary files, you can't really do that.

It would be better to find some sort of precondition that we could add that would limit this to the servers in question, and stick with one file, I would think.

Or see my previous comment, and not need to do any cleanup in liquibase at all...

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mogoodrich: @mseaton  yeah, that was my original idea, but then I realized that wouldn't work, because if a new location was added (or any change was made) to locations.csv, on the next restart all the tags would be wiped again when locations.csv was re-processed... and unless changes were also made to the locationtagmaps, the tag files would not get reprocessed.

api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
mogoodrich: Okay, I actually did notice today a precondition that Ellen used in the main liquibase file (checking the address hierarchy for a location "Haiti") that I could use instead.  That would negate the need for this change set entirely.

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mseaton: @mogoodrich I wonder if we want to configure the locationtagmaps domain to run "always", and ignore checksum.  Would that solve the issue?

mogoodrich: If that's a possible configuration, then, yeah, that would work... I think actually we'd want "always" for most metadata (except those that are time-consuming to load)...

mseaton: Looking at the docs, it appears you can specify a runtime property that indicates to skip checksums globally (and reload everything), but not on a domain-by-domain basis...

mogoodrich:
I reverted the changes to load multiple liquibase files, but kept the change to move the liquibase call slightly earlier in the build.

mseaton (approved):

PR #308, opened 2021-09-21
Uhm 5420

mseaton:


mseaton:
This PR adds DataSetManager classes to pihcore for those parts of the FullDataExportBuilder that hard-code dataset creation in conditionals within that class, and adds these alongside the other datasetmanagers that contribute to the full data export in the pihcore module.

PR #311, opened 2021-09-27
UHM-5400, add encounterList state to list encounters outside the visits

cioan:
I just added another encounterList state view to the visit controller. This allows us to list all encounters, filter by encounterType that are outside of a visit.


mogoodrich (commented):
Thanks @cioan !  Perhaps I'm not following 100%, but is this supposed to be part of the Visit view? If not, would it make sense to make it a entirely different component, instead of part of visit.js?

cioan:
I have started building it as a separate component, but then I realized that there is so much easier to just add another view to the existing Angular component. So, we could use a lot of the functionality that has been built into the visit controller. Otherwise, I am afraid we would have to copy and paste some code, or maybe extract some common functions to other components, but I am not sure at this point if this type of components(gsp + Angular) have a future in our development roadmap?

mogoodrich:
Fair enough, @cioan ... maybe we can add this in but tweak the documentation (or even the name of the Visit controller)... let's merge this and then perhaps do a post-commit review... I think it will be easier for me to review once I have a look at how it works.

PR #312, opened 2021-09-28
UHM-5400, add edit and delete encounter icons

cioan:
Just added two icons next to each encounter row that allows us to edit and delete an encounter.


mogoodrich (approved):
LGTM, thanks @cioan !

PR #313, opened 2021-09-30
Uhm 5823

mseaton:


mogoodrich (commented):
Pretty cool, and generally looks good to me, let's discuss after standup.

mseaton:
@mogoodrich - do you want to discuss this before merging?  My biggest question is around the actual class names and modeling, but assuming we don't run off and build too many dependencies on this right away, we can always evolve it over time.  Happy to discuss early next week, but don't want to leave it too long.

mogoodrich:
I'm fine with merging and evolving...

PR #317, opened 2021-10-07
UHM-5836: Add late med pickup, late to appointment on HIV Clinical Da…

mogoodrich:
…shboard (work-in-progress)

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/HivAlertsFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/HivAlertsFragmentController.java
@@ -1,0 +1,97 @@
+package org.openmrs.module.pihcore.fragment.controller.dashboardwidgets;
+
+import org.openmrs.Concept;
+import org.openmrs.Encounter;
+import org.openmrs.EncounterType;
+import org.openmrs.Obs;
+import org.openmrs.Patient;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.EncounterService;
+import org.openmrs.api.ObsService;
+import org.openmrs.api.PatientService;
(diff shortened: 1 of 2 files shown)


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/HivAlertsFragmentController.java
```diff
@@ -0,0 +1,147 @@
...
+				throw new IllegalArgumentException("Patient must be of type Patient or PatientDomainWrapper");
+			}
+		}
+
+		generateLateForMedPickupAlert(model, patientWrapper, encounterService, conceptService, obsService);
+		generateLateForAppointmentAlert(model, patientWrapper, encounterService, conceptService,obsService);
+
+	}
+
+	private void generateLateForMedPickupAlert(FragmentModel model, PatientDomainWrapper patientWrapper, EncounterService encounterService, ConceptService conceptService, ObsService obsService) {
```
mseaton, line 68: I won't pick through all of this logic yet.  My main thought is that this logic should not be hardcoded in here like this in a way that will require programming to change and which is opaque to the clinical and implementation team as to how it is calculated.  

A challenge with the HIV EMR is often that the definition for things like "lost to followup", "late for visit", "late for med pickup", etc. change very frequently as protocols or CDC/Pepfar definitions change.  At the same time, for every alert on the dashboard, there is likely a corresponding set of other tools that need the same criteria.  These include:

* Give me a data export of all patients who are currently late for their med pickup
* Build me an indicator around the number of patients late for their med pickup
* Give me an interactive patient list that shows me all patients who are late for their appointment

I wonder if an alternative here would be to have a set of report or data set definitions, defined via configuration, that this alert controller could iterate over and use to determine which alerts to display for a given patient.

Thoughts @mogoodrich ?

PR #318, opened 2021-10-14
UHM-5817: validate format of Haiti phone number

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/config/registration/ContactPersonConfigDescriptor.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class ContactPersonConfigDescriptor {
+
+	@JsonProperty
+	private Boolean required;
+
+	@JsonProperty
+	private RegistrationFieldDescriptor phoneNumber;
+
+	public Boolean getRequired() {
```
mogoodrich, line 13: I decided to extract the "required" configuration out to the PIH Config as well while I was at it.

PR #320, opened 2021-11-05
Localized formatting of dates in the visit note

brandones:
Depends on https://github.com/openmrs/openmrs-module-uicommons/pull/98

English:
Screenshot from 2021-11-05 16-06-05

Spanish:
Screenshot from 2021-11-05 16-05-46

French:
Screenshot from 2021-11-05 16-06-20

Still displays in English for Haitian, but that's not actually a regression, because these have always displayed in English. They'll start displaying in French format next year, pending a Unicode update.

### DIFF
--- a/omod/src/main/webapp/resources/scripts/visit/constants.js
+++ b/omod/src/main/webapp/resources/scripts/visit/constants.js
@@ -1,7 +1,12 @@
 angular.module('constants', [])
     .value('DatetimeFormats', {
         date: "d-MMM-yy",
+        dateLocalized: { year: 'numeric', month: 'short', day: 'numeric' },
         time: "hh:mm a",
+        // TODO: fix timeLocalized:
+        //   'hour12' override is necessary because we're using `fr` in Haiti instead of `fr_HT`.
+        //   This override will prevent time from displaying correctly in places that use a 24 hour clock.
+        timeLocalized: { hour: '2-digit', minute: '2-digit', hour12: true },
         datetime: "d-MMM-yy (hh:mm a)"
     })
 
--- a/omod/src/main/webapp/pages/visit/templates/encounters/defaultEncounterHeading.gsp
+++ b/omod/src/main/webapp/pages/visit/templates/encounters/defaultEncounterHeading.gsp
@@ -4,10 +4,10 @@
     </div>
     <div class="col-4 text-right">
         <span class="date-span">
-            <i class="icon-calendar"></i>{{ encounter.encounterDatetime | serverDate:DatetimeFormats.date }}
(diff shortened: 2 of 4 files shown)


mseaton (approved):
LGTM - assuming that this is what everyone wants in all of the affected countries

mogoodrich:
@brandones  @mseaton  what was the existing date localization problem we were overcoming here?

omod/src/main/webapp/resources/scripts/visit/constants.js
```diff
@@ -1,7 +1,12 @@
 angular.module('constants', [])
     .value('DatetimeFormats', {
         date: "d-MMM-yy",
+        dateLocalized: { year: 'numeric', month: 'short', day: 'numeric' },
         time: "hh:mm a",
+        // TODO: fix timeLocalized:
+        //   'hour12' override is necessary because we're using `fr` in Haiti instead of `fr_HT`.
+        //   This override will prevent time from displaying correctly in places that use a 24 hour clock.
```
mogoodrich, line 8: Confused by this comment... it sounds like this override is meant to prevent time from being display correctly, which I assume is not the case? @brandones  @mseaton ?

brandones: We always display 12-hour time (AM/PM). This is okay in Haiti, Mexico, Peru, Sierra Leone, and Liberia; but it is incorrect in Rwanda, Lesotho, and Kazakhstan, which all use the 24-hour clock. Fortunately the PIH EMR isn't used in those latter countries.

Normally if you use the JavaScript date/time localization functions they will render in 12 or 24 hour format depending on the locale. However, `fr` uses the 24-hour clock (as in France), but we use `fr` in Haiti instead of `fr_HT`. By setting `hour12: true`, we force it to show 12-hour time, even for the `fr` locale.

mogoodrich:
So this needs further review... it corrects the problem where the date displays aren't localized into French, Kreyol and Spanish, but it introduces new issues and inconsistencies when the time zone of the client doesn't match the time zone of the server.  

he's a screenshot of a patient, viewed in the French locale, with time zone setting on my laptop set to a time *earlier* than server time (EST):

Note that this is retrospectively created visit and encounter on Nov 9th.  Note that it is a visit on Nov 9th, with a single encounter at 12:00am (which generally represents that we don't know the time of the encounter).  Note that although the locale is in French, the time zone is not localized.



Here's after the change.  Note that the date is now properly displayed in French, but the visit date ranges from Nov 6th and Nov 7th (due to the the time zone shift), but the encounter datetime is still for Nov 9th at midnight (becaues assumedly that template still uses the old "serverDate" filter, while the one that displays the visit date does not).



More confusingly, if I open the visit date for editing, the "current" value does not match the default value in the date select...



.... and then if I save without changing the change, the visit date shifts a day earlier...

PR #321, opened 2021-11-09
Only use the language code for Angular locale

brandones:


mseaton (commented):
Why remove the country / variant parts of the locale?  I'm confused.

brandones:
@mseaton The problem I brought up here: https://github.com/PIH/openmrs-config-ses/pull/46#issuecomment-963662665

mogoodrich:
@brandones  @mseaton  this worked with the current functionality, which only supports language anyway, but don't we need to build in support for country to support Peru vs Mexico translations?  Or is the issue that there are no different translations needed for the Visit Note?  (If that's the case, I'm fine with this, but we should ticket/remember we will need to expand the Message Controller to support count custom-by-country if we need this in the future)

omod/src/main/java/org/openmrs/module/pihcore/page/controller/visit/VisitPageController.java
```diff
@@ -66,7 +66,7 @@ else if (encounter != null) {
         model.addAttribute("encounterType", encounterType);
         model.addAttribute("suppressActions", (suppressActions != null) ? suppressActions : false);
         model.addAttribute("encounter", encounter);
-        model.addAttribute("locale", uiSessionContext.getLocale());
+        model.addAttribute("locale", uiSessionContext.getLocale().getLanguage());
```
mogoodrich, line 69: Can we add a in-line comment with the reasoning behind this (MessagesController currently only supports language?)

mseaton: Right.  I understand the issue you are facing, I just don't understand why the solution is to remove country here, rather than to add country support into the Message Controller.  Isn't our goal to support proper localization?

brandones: Just got off a call with @mogoodrich who recalls that the last time he tried that, it introduced other bugs. It is, of course, the right thing to do. Can I ticket it and merge this in, @mseaton ? Would be great to get Peru as un-blocked as possible.

mogoodrich: works for me...

brandones:
Created a ticket to add country support to the message controller: https://pihemr.atlassian.net/browse/UHM-5926

omod/src/main/java/org/openmrs/module/pihcore/page/controller/visit/VisitPageController.java
mseaton: Sounds good.

PR #322, opened 2021-11-09
Improve behavior for fetching historical encounters in visits app

brandones:
Before, the UI would sometimes fail to update after a state change. Now it seems to work consistently.

Follow-up to https://github.com/PIH/openmrs-module-pihcore/pull/319
Related to https://github.com/openmrs/openmrs-module-coreapps/pull/438
Ticket https://pihemr.atlassian.net/browse/UHM-5851


mogoodrich (approved):
Generally looks good to me, though I'm not an expert in Angular scope refresh!

PR #324, opened 2021-11-10
Use the serverDateLocalized function to render dates and times

brandones:
Depends on https://github.com/openmrs/openmrs-module-uicommons/pull/99


brandones:
@mogoodrich is there an easy way to test whether this reproduces the behavior you expect?

mogoodrich (approved):
LGTM, but additionally we should do a global search-and-replace of "serverDate:" with 'serverDateLocalized" as there are several other "heading" gsps following the same pattern that use serverDate.

PR #325, opened 2021-11-15
Add peru prescription component

MiguelAHPpih:


api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -67,6 +67,7 @@
         public static final String VISIT_NOTE = "visitNote";  // get rid of the left visits bar and enable form sections
         public static final String ALLERGIES = "allergies";
         public static final String PRESCRIPTIONS = "prescriptions";  // TODO no longer used?
+        public static final String PERU_PRESCRIPTIONS="peruPrescriptions";
```
mogoodrich, line 70: I think we could just name this "Prescription":

mogoodrich (commented):
LGTM, with the general suggestion that I think we can just name this "Prescription"

api/src/main/java/org/openmrs/module/pihcore/config/Components.java
```diff
@@ -67,6 +67,7 @@
         public static final String VISIT_NOTE = "visitNote";  // get rid of the left visits bar and enable form sections
         public static final String ALLERGIES = "allergies";
         public static final String PRESCRIPTIONS = "prescriptions";  // TODO no longer used?
+        public static final String PRESCRIPTION="prescription"; //Used for Perú
```
brandones, line 70: I hadn't noticed that we have an old component named `prescriptions`. We should clean that up.

brandones (approved):

api/src/main/java/org/openmrs/module/pihcore/config/Components.java
MiguelAHPpih: Do you want me to erase it or make into a comment?

brandones: Oh, you don't need to do anything. We'd have to make sure it's not used anywhere, etc. I was just making a note to myself and @mogoodrich .

mogoodrich: Ah, good catch @brandones ... I just went ahead and removed this... if we are wrong, and this is still used, I'd rather it fail fast and clean it up... having both a "prescription" and a "prescriptions" component would be confusing.

PR #326, opened 2021-11-17
Adds a "Unit Test" that can be executed to update the message propert…

mseaton:
…ies files to match the ordering and organization in the primary english file, and comment out any properties that are unspecified or which match the english exactly.


mogoodrich (approved):

brandones:
So this is intended to be run locally by developers, who can then commit the results?

Seems reasonable to me.

PR #329, opened 2021-11-30
UHM-5966 Remove metadatadeploy module from distro

mseaton:
Remove metadatadeploy as a dependency of pihcore


omod/pom.xml
```diff
@@ -171,7 +171,7 @@
 							<includeGroupIds>${project.parent.groupId}</includeGroupIds>
 							<includeArtifactIds>${project.parent.artifactId}-api</includeArtifactIds>
 							<excludeTransitive>true</excludeTransitive>
-							<includes>**/*</includes>
+							<includes>moduleApplicationContext.xml,liquibase.xml</includes>
```
mseaton, line 174: @mogoodrich - this is the area in question.  The change I made basically says - we want to copy these specific files into the omod from the api project, whereas before it was casting a wide net and somehow also picking up all of the test classes and resources (maybe the "includeArtifactIds" includes anything that starts with the id in question, so it would also pick up artifact pihcore-api-tests ?

mogoodrich: Strange, nice catch... I see this is in Core Apps as well and dates back to at least 2013.

do we have an idea of what this is supposed to do?  The id says "Expand moduleApplicationContext and messages"?

Assumedly we need  Talk post about this? (Maybe you've already done so, I'm behind)

mogoodrich (commented):
Generally looks good, though probably easier if you walk me through the pom change, which, yes, it sounds like we should address sooner rather than later.

omod/pom.xml
mogoodrich: Actually, if I'm understanding the problem right, I'm only seeing this manifest itself in the PIH Code module, but maybe I'm misunderstanding?  Let's talk after standup?

mseaton: Let's take the bigger picture question of where we might need to fix this in the OpenMRS ecosystem separately from this specific change for this specific PR, and not hold this up over the bigger picture question.

mogoodrich: Sure, the issue now is that I'm only seeing this problem manifest itself in PIH Core, and this code is used everywhere, so I'm wondering if something else is wrong with PIH Core?  (but again, I could be misunderstanding the issue)

PR #338, opened 2021-12-08
UHM-4342 - Provide ability to setup PIH EMR environment utilizing sys…

mseaton:
…tem properties.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/PihCoreUtil.java
@@ -1,5 +1,6 @@
 package org.openmrs.module.pihcore;
 
+import org.apache.commons.lang.StringUtils;
 import org.openmrs.Visit;
 import org.openmrs.api.context.Context;
 import org.openmrs.util.OpenmrsUtil;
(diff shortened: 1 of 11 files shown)


mseaton:
This PR makes 3 main changes to how we configure pihcore:

1. It changes the "local haiti identifier generator" configuration from a runtime property to something setup in pih.config, so this is no longer a manual step that must be taken, but something that is specific to the type of server you are trying to install (i.e. all "test/dev" servers for mirebalais/haiti should be set up with a local identifier source, and all production servers should not.

2. It uses the new feature in Initializer to enable setting the startup.load=disabled property via a system property as an alternative to a runtime property, and uses the built-in SDK feature to ensure this system property is set for all PIHEMR servers.

3. It updates the pihcore configuration to support reading the pih.config and pih.config.directory properties from either a System property or a Runtime property, and then uses the built-in SDK feature to prompt users at the time of doing "setup" what PIH Config values they wish to set up their server with, and then configures these to be present as system properties for that server.

The result is that (when applied to changes to the distro - see related PR in openmrs-config-distro), the PIH EMR can be setup and run without many of the workarounds we currently have in place to seed the runtime properties file with the correct values before the pihcore module first starts up.

mogoodrich (approved):
LGTM, thanks @mseaton

PR #340, opened 2021-12-15
change parameters so that PIH Core "pihcore.started" doesn't get set …

mogoodrich:
…to "false" on failed startup


brandones (approved):
Assuming you tested both a normal restart and a restart after a pihcore failure, LGTM

mseaton (approved):
I'll take your word for it.

PR #342, opened 2021-12-20
PER-406: Unable to set up new Peru dev environment from scratch

mogoodrich:
(Add Concepts to the list of metadata to be loaded *after* concepts)


api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java
```diff
@@ -27,6 +28,7 @@
 
     public static List<Domain> getDomainsToLoadAfterConcepts() {
         return Arrays.asList(
+				CONCEPTS,
```
brandones, line 31:
```
                CONCEPTS,
```

brandones (approved):
Try to keep files consistent between tabs and spaces

brandones:
LGTM thanks!

PR #343, opened 2021-12-28
UHM-5925 - Remove MirebalaisMessageSource in favor of InitializerMess…

mseaton:
…ageSource


mogoodrich (commented):
Looks pretty straightforward, just substituting the custom message source for the Iniz one.

I did see some failing tests for the PIH Core module, but assumedly that was due to the latest Iniz not being appropriately deployed and it looks like you took care of it.

PR #349, opened 2022-02-01
UHM-6145 Allow multiple obgyn encounters per visit

lnball:
Quick review.  There are more complicated ways to handle this, but this is a short term fix as requested by ZL.


mogoodrich (approved):
LGTM!

PR #350, opened 2022-02-02
UHM-6149 - Only include report when site is unspecified only if count…

mseaton:
…ry is also unspecified


jmbabazi:
LGTM! thanks @mseaton

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -1005,7 +1005,7 @@ private void enableReportsAndExports() {
 
                 boolean matchesComponent = (components == null || config.anyComponentEnabled(components));
                 boolean matchesCountry = (countries == null || countries.contains(config.getCountry().name()));
-                boolean matchesSite = (sites == null || sites.contains(config.getSite()));
+                boolean matchesSite = ((countries == null && sites == null) || sites.contains(config.getSite()));
```
mogoodrich, line 1008: LGTM!

PR #354, opened 2022-02-11
UHM-6037 - Add Terms of Use acknowledgement

mseaton:


omod/src/main/java/org/openmrs/module/pihcore/page/controller/account/TermsAndConditionsPageController.java
```diff
@@ -0,0 +1,47 @@
...
+    public String post(@RequestParam(value = "termsAndConditionsAccepted", defaultValue = "false") boolean termsAndConditionsAccepted,
+                       @SpringBean("userService") UserService userService) {
+
+        if (termsAndConditionsAccepted) {
+            User currentUser = Context.getAuthenticatedUser();
+            String acceptanceDate = new ISO8601DateFormat().format(new Date());
+            currentUser.setUserProperty(PihCoreConstants.USER_PROPERTY_TERMS_AND_CONDITIONS_ACCEPTED_DATE, acceptanceDate);
+            userService.saveUser(currentUser);
+        }
+        return "redirect:/pihcore/home.page";
```
mogoodrich, line 45: Assumedly if the terms and conditions aren't accepted, the filter handles the redirect/block when hitting the home page?  (ie, this seems to redirect to the home page whether or not the the terms are accepted)

omod/src/main/java/org/openmrs/module/pihcore/TermsAndConditionsFilter.java
```diff
@@ -0,0 +1,111 @@
...
+			HttpServletRequest httpRequest = (HttpServletRequest) request;
+			try {
+				// If we have not already indicated on the session that the Terms were accepted, run the check
+				if (!"true".equals(httpRequest.getSession().getAttribute(SESSION_ATTRIBUTE_TERMS_ACCEPTED))) {
+					User currentUser = Context.getAuthenticatedUser();
+					// Only check once a user has logged in successfully, so we can check and update their account
+					if (currentUser != null) {
+						String propertyVal = currentUser.getUserProperty(PihCoreConstants.USER_PROPERTY_TERMS_AND_CONDITIONS_ACCEPTED_DATE);
+						// If the user had previously accepted the terms, update the session with this information
+						if (StringUtils.isNotBlank(propertyVal)) {
```
mogoodrich, line 74: The assumption here is that there's never a "false" state for this property (makes sense to me)

mogoodrich (approved):
LGTM, thanks @mseaton !

omod/src/main/java/org/openmrs/module/pihcore/TermsAndConditionsFilter.java
mogoodrich: Makes sense, thanks!

PR #357, opened 2022-02-18
UHM-6194 Remove org.openmrs.module.mirebalaisreport mappings and source

lnball:
See ticket for more information.  This is messy with updates in 2 modules and some issues.


mseaton (commented):
@lnball see my comments on the ticket, but this is just test data that is only loaded for unit testing.  It will never make it onto any system.  Note that both of the files you changed are in the “src/test/resources” folder.  The reason why things fail when you made the changes you did, is because you removed the concept_reference_source entries, but did not remove all of the concept_reference_terms that refer to these sources.  Search for _concept_source_id="11"_ and you'll see what I mean.

I recommend we close this PR and not merge it in.

PR #359, opened 2022-02-26
UHM-6023 - Add "forgot password" option

mseaton:


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/account/ResetPasswordFragmentController.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * I don't know why this is implemented this way in OpenMRS core, but this is needed to get things to work
+     */
+    private synchronized void ensureHostUrl(AdministrationService administrationService, HttpServletRequest request) {
+        try {
+            Context.addProxyPrivilege(PrivilegeConstants.MANAGE_GLOBAL_PROPERTIES);
+            String requestUrl = request.getRequestURL().toString();
+            String hostUrl = requestUrl.replace("resetPassword/reset.action", "resetPassword.page");
+            hostUrl += "?activationKey={activationKey}";
+            administrationService.setGlobalProperty(OpenmrsConstants.GP_HOST_URL, hostUrl);
```
mogoodrich, line 63: This is bizarre, not sure I follow?  Does core blow this GP away or something each time?

I'm not really following what is happening, but my main concern here is any issues that might occur if multiple password resets happen at the same time? No way for the wrong activation key to be sent to a user?

omod/src/main/webapp/pages/account/resetPassword.gsp
```diff
@@ -0,0 +1,57 @@
+<!-- This is the login page -->
```
mogoodrich, line 1: Is it?

omod/src/main/webapp/pages/account/resetPassword.gsp
```diff
@@ -0,0 +1,57 @@
...
+    <link rel="shortcut icon" type="image/ico" href="/${ ui.contextPath() }/images/openmrs-favicon.ico"/>
+    <link rel="icon" type="image/png\" href="/${ ui.contextPath() }/images/openmrs-favicon.png"/>
+    ${ ui.resourceLinks() }
+</head>
+<body>
+<script type="text/javascript">
+    var OPENMRS_CONTEXT_PATH = '${ ui.contextPath() }';
+    var errorMessageOldPassword = "${ui.message("emr.account.changePassword.oldPassword.required")}";
+    var errorMessageNewPassword = "${ui.message("emr.account.changePassword.newPassword.required")}";
+    var errorMessageNewAndConfirmPassword = "${ui.message("emr.account.changePassword.newAndConfirmPassword.DoesNotMatch")}";
```
mogoodrich, line 21: Are these used?

mogoodrich: So these are used?  Seems not sure why the old password would be required on this page since it isn't known?

PR #361, opened 2022-03-02
UHM-6250 - Password reset email and page should be localized to curre…

mseaton:
…nt user, even if not authenticated

Note:  This is dependent on API changes to core in TRUNK-6069


mogoodrich (approved):
LGTM, thanks @mseaton

PR #364, opened 2022-03-08
UHM-6136: View/Edit page for saved forms

mogoodrich:


cioan (approved):
Looks good, @mogoodrich ! As a potential alternative to passing query parameters in the URL, maybe we could store the page state in the browser local store?

brandones:
Is the goal to be able to navigate to the page in one of a variety of states (in which case, use query params), or is the goal to persist the state across user visits to the page (in which case, use localStorage)?

mogoodrich:
@brandones  more the former... the Visit (angular) app now has a specific view that hides all other information except the selected encounter, which we use when navigating from the encounter list page.  When a user edits a section of a form (which navigates away from the page into the HFE UI) we want the page to still load in the "encounterOverview" state... so we want it to persist, but only for that specific workflow, not across all visits.

mogoodrich:
Thanks @cioan  and @brandones .... going to merge... also, I'll think more about how we could use local storage or session storage as perhaps a better way to do this if this doesn't fix things....

PR #371, opened 2022-03-22
LIB-63 Added new name for Mental health FU action/form

lnball:


mogoodrich (commented):
@lnball, this works, but you should be able to just override "pih.task.mentalHealth.label" in the new messages_liberia_en.properties you created as well.

PR #372, opened 2022-03-23
UHM-6330: View/Edit page for saved forms -- Forms that aren't a part …

mogoodrich:
…of a visit

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/page/controller/patient/EncounterListPageController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/page/controller/patient/EncounterListPageController.java
@@ -19,7 +19,7 @@ public class EncounterListPageController {
 
     public void get(PageModel model, UiUtils ui,
                       @InjectBeans PatientDomainWrapper patientDomainWrapper,
-                      @RequestParam(value = "patient") Patient patient,
+                      @RequestParam(value = "patientId") Patient patient,
                       @SpringBean("encounterService") EncounterService encounterService) throws IOException {
 
         // Get all of the encounters for the patient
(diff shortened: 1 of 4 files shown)


omod/src/main/java/org/openmrs/module/pihcore/page/controller/patient/EncounterListPageController.java
```diff
@@ -19,7 +19,7 @@
 
     public void get(PageModel model, UiUtils ui,
                       @InjectBeans PatientDomainWrapper patientDomainWrapper,
-                      @RequestParam(value = "patient") Patient patient,
```
mogoodrich, line 22: HFE-UI automatically applies the patient id as a "patientId" parameter on any return url it generates, so adapting this to work with that.  (Would have been better to call it "patient", but it is what it is... the underlying mapper can handle both ids and uuids).

omod/src/main/webapp/pages/patient/encounterList.gsp
```diff
@@ -1,6 +1,15 @@
 <%
     ui.decorateWith("appui", "standardEmrPage")
     ui.includeJavascript("uicommons", "datatables/jquery.dataTables.min.js")
+
+    // hack, hardcode what encounter types use the simple view
```
mogoodrich, line 5: Obviously a hack... we could consider way to make this configurable, but I wasn't sure if it was worth the effort, because it seemed like it would just be a hacky way to make it configurable (ie gp: "pihcore.encounterTypesRenderedWithSimpleView")?  At least here it's hack, but not propagated and use is limited to a single file.

mseaton (approved):
Looks good to me @mogoodrich

mogoodrich:
Thanks @mseaton , merging!

PR #377, opened 2022-04-15
UHM-6390: Pilot Html Forms in O3.  This ensures all forms are publish…

mseaton:
…ed.  This also fixes a false error in the logs in systems (eg. Mexico) that does not have a drugs csv.


mogoodrich (approved):
LGTM.  I guess we should keep an eye out for any ramifications of publishing our forms, but,yeah are forms certainly are "published", so this seems to be the correct state.

brandones:
It would make sense to me that there would be some mechanism for controlling whether a form is published or not. And then publication status would be an effective way for the backend to tell the frontend which forms should actually appear.

mseaton:
Yeah, though often what forms are available is not a simple on/off setting.

brandones:
You mean with respect to permissions, patient age and gender, etc? That kind of thing I am inclined to handle on the frontend since we don't have any way to handle it on the backend.

PR #378, opened 2022-04-21
2.x

mseaton:


api/src/main/java/org/openmrs/module/pihcore/mpi/ImportPatientFromWebService.java
```diff
@@ -0,0 +1,328 @@
...
+    private Date parseDate(String date) {
+        try {
+            DateFormat dateFormat = new SimpleDateFormat(FULL_DATE_FORMAT);
+            return dateFormat.parse(date);
+        } catch (ParseException e) {
+            throw new IllegalArgumentException("Badly formatted date: " + date);
+        }
+    }
+
+}
```
mogoodrich, line 328: Assuming this is just moved wholesale from ImportFromWS, if there's anything in particular you think I should look at here, LMK...

api/src/test/java/org/openmrs/module/pihcore/ConceptDictionaryExportTest.java
```diff
@@ -40,7 +40,7 @@
  * mvn test -Dtest=ConceptDictionaryExportTest -Dconcept_export_properties_file=/tmp/concept_export.properties
  */
 @SkipBaseSetup
-@Ignore
+@Disabled
```
mogoodrich, line 43: The new annotation for ignore, I assume?

mogoodrich (approved):
Sweet, thanks so much for slogging through this @mseaton ... like you said, tons of changes, but mainly just a bunch of similar changes to the tests, but looks good.

I did see bunch of changes from @Before -> @BeforeEach, which were two annotations with different means previously (run once before running all the tests vs. running before each test, but I assume you are aware of that and there's a reason for this)

api/src/test/java/org/openmrs/module/pihcore/ConceptDictionaryExportTest.java
mseaton: Seems that way in Junit 5 yes

mogoodrich (changes requested):
Let's update the maven versions plugin in the main pom.xml here as well.
You can we what we currently have it set to:
[2.3.4-SNAPSHOT,2.3.4]
Could likely be something as simple as [2.5.5-SNAPSHOT,2.5.5] which should mean that it will update to 2.5.5 when it is released, but nothing beyond that.

mseaton:
Updated @mogoodrich - see latest commit:  f4587c9a359d1f2ab4f3e22c32409dac73d54def

PR #379, opened 2022-04-25
UHM-6363: Configure (and release) link from local EMRs to HIV Cloud

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -3033,6 +3039,20 @@ private void enableCohortBuilder() {
 
     }
 
+    private void enableHaitiHIVLink() {
+        extensions.add(extension(
+                CustomAppLoaderConstants.Extensions.HAITI_HIV_EMR_DASHBOARD_LINK,
+                "pihcore.hiv.viewPatientInHivEmr",
+                "icon-external-link",
+                "link",
+                "https://haitihivtest.pih-emr.org/openmrs/pihcore/patient.page?patientId={{identifier}}&identifierType=" + ZlConfigConstants.PATIENTIDENTIFIERTYPE_HIVEMRV1_UUID + "&dashboard="+ PihEmrConfigConstants.PROGRAM_HIV_UUID,
```
mogoodrich, line 3048: I'll test that this works with the "haitihivtest" server on HUM-CI, and then change this to "haiti" and disable on HUM-CI.

mseaton, line 3048: I don't think we should have actual server names in any of our code, including configs, if we can avoid it.  Both for security reasons and also in case this needs to change.  These should come from Puppet and should be able to be encrypted in github.  So whatever server url is configured here needs to be something you are reading from a runtime property that is populated from Puppet I think.

mseaton (changes requested):
See comments.

mogoodrich:
Not sure how I messed up that commit message, will fix on merge.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
mogoodrich: Updated to use runtime property.  Note that since this needs to be configured, I removed the component and just based the feature on whether or not the runtime property is present.

Will follow up with the updates to Mirebalais Puppet tomorrow.

mseaton (approved):

mogoodrich:
Thanks @mseaton !

PR #382, opened 2022-04-26
UHM-6440: Remove Check-In and Vital Signs app from home page

mogoodrich:


mogoodrich:
Note this is only about removing the Vitals Home Page app... the Check-In Homepage app already had a separate component, so it was trivial to remove.

cioan (approved):
LGTM!

PR #383, opened 2022-04-28
throw error and stop startup if liquibase update fails

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
```diff
@@ -3,17 +3,19 @@
...
     protected static Log log = LogFactory.getLog(LiquibaseSetup.class);
 
-    public static void setup() {
+    public static void setup() throws Exception {
         try {
             DatabaseUpdater.executeChangelog(PihCoreUtil.getLiquibaseChangeLog(), null);
         } catch (Exception e) {
-            log.error("Unable run liquibase change sets", e);
+            log.error("Unable run liquibase change sets provided by PIH EMR config", e);
+            throw new DatabaseUpdateException(e);
```
mseaton, line 18: Why not just "throw e"?

mogoodrich: Good point... I updated (and will merge now)

PR #384, opened 2022-05-04
UHM-5470: CloseStaleVisitsTask failing on HSN server

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
```diff
@@ -53,6 +53,10 @@ public void run() {
         LocationTag visitLocationTag =  locationService.getLocationTagByName(EmrApiConstants.LOCATION_TAG_SUPPORTS_VISITS);
         List<Location> locations = locationService.getLocationsByTag(visitLocationTag);
 
+        if (locations == null || locations.size() == 0) {
+            log.error("Unable to close stale visits, no locations with Visit Location tag");
```
mseaton, line 57: Not sure why there are systems out there with no "Visit Locations" configured?  But this seems fine.  If this is not "error" behavior, then I'd probably change this to a log.debug unless you want to fill the logs with this message every 5 minutes.

PR #392, opened 2022-07-26
UHM-6612: Analyze (and hopefully fix) issues with running PIH EMR aga…

mogoodrich:
…in OpenMRS 2.6.x (current master)


mseaton (approved):
OK, I guess.  Why just here on login?

mogoodrich:
It was added to the standard emr pages here by Daniel, which we evidently don't use for the login page... if we don't have a ticket about cleaning up/fixing the login page to use something more standard, we probably should:

https://github.com/openmrs/openmrs-module-appui/commit/3d64ab6c6a55c480f81995b03567541c51a6ed0f

PR #397, opened 2022-09-06
UHM-6696: PIH EMR not automatically upgrading OpenMRS versions after …

mogoodrich:
…upgrade to new Bamboo

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,11 +62,11 @@
         <emrVersion>2.3.0-SNAPSHOT</emrVersion>
         <emrapiVersion>1.33.0</emrapiVersion>
         <eventVersion>2.10.0</eventVersion>
-        <fhir2Version>1.4.0-SNAPSHOT</fhir2Version>
+        <fhir2Version>1.5.0</fhir2Version>
+        <growthChartVersion>2.1.0</growthChartVersion>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -440,6 +504,20 @@
...
         </dependency>
 
+        <dependency>
+            <groupId>org.openmrs.module</groupId>
+            <artifactId>uilibrary-api</artifactId>
+            <version>${uilibraryVersion}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
```
mseaton, line 514: No, this cannot go here.  This must be moved into the omod project, unless we use a dependencyManagement section (which we don't have).  The api cannot contain omod artifacts.

mseaton (changes requested):

pom.xml
mogoodrich: Fair enough... will do a little testing to see if this does what is expected when it's in the omod, but otherwise we can just punt on the REST updates for now.

mogoodrich: Looks like it doesn't pick up the REST web services version upgrade if it's not referenced in that pom... will just not worry about that for now, we can upgrade REST web services manually as need be for now.

PR #398, opened 2022-09-07
UM-6709: Support Custom liquibase files for countries

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
```diff
@@ -6,16 +6,26 @@
...
 public class LiquibaseSetup {
 
     protected static Log log = LogFactory.getLog(LiquibaseSetup.class);
 
-    public static void setup() throws Exception {
+    public static void setup(Config config) throws Exception {
         try {
             updateLiquibaseChangeLogPath();
             DatabaseUpdater.executeChangelog(PihCoreUtil.getLiquibaseChangeLog(), (ChangeSetExecutorCallback) null);
+            if (new File(OpenmrsUtil.getApplicationDataDirectory()  + "/" + PihCoreUtil.getSiteSpecificChangeLog(config)).exists()) {
```
mseaton, line 26: I would think this should be new "File(OpenmrsUtil.getApplicationDataDirectory(), PihCoreUtil.getSiteSpecificChangeLog(config))" - eg. just separate the dir and file with a comma as separate arguments.

api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java
```diff
@@ -6,16 +6,26 @@
...
+                DatabaseUpdater.executeChangelog(PihCoreUtil.getSiteSpecificChangeLog(config), (ChangeSetExecutorCallback) null);
```
mseaton, line 27: Would be better to call PihCoreUtil.getSiteSpecificChangeLog(config) once above, and then use it in both places

mseaton (approved):
Looks generally fine, but a few style suggestions

PR #401, opened 2022-10-06
UHM-6728 - Activity Log no longer logging after upgrade to 2.5.x

mseaton:


omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
```diff
@@ -3,26 +3,19 @@
  * Version 1.0 (the "License"); you may not use this file except in
  * compliance with the License. You may obtain a copy of the License at
  * http://license.openmrs.org
- *
+ * <p>
```
mogoodrich, line 6: Strange auto-html formatting? or did the license template change?

omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
```diff
@@ -3,26 +3,19 @@
...
  * Software distributed under the License is distributed on an "AS IS"
  * basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
  * License for the specific language governing rights and limitations
  * under the License.
- *
+ * <p>
```
mogoodrich, line 11: and here

omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
```diff
@@ -76,53 +44,29 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+				try {
+					HttpServletRequest httpRequest = (HttpServletRequest) request;
+					HttpSession session = httpRequest.getSession();
+					ThreadContext.put("sessionId", session.getId());
+					ThreadContext.put("user", (String) session.getAttribute("username"));
+					ThreadContext.put("loadTime", Long.toString(totalTime));
+					ThreadContext.put("method", httpRequest.getMethod());
+					ThreadContext.put("ipAddress", httpRequest.getRemoteAddr());
+					ThreadContext.put("servletPath", httpRequest.getServletPath());
+					ThreadContext.put("requestURI", httpRequest.getRequestURI());
```
mogoodrich, line 58: How does this thread context stuff work? does it get logged by the next "logger.trace" line?

mogoodrich (approved):
I don't quite follow the format of the output, do you have an example?

If the DataKind project does make headway, I think there will be a desire to have the new activity log format closely match the existing format, as they will likely have written some targeted scripts for slurping the logs.

I also think that we would need way to not have this stuff go to catalina.out, because I think it will make catalina.out bloated and hard to follow (or harder to follow than it currently is :)

*BUT* that all being said, all that can be "we will cross that bridge if/when we come to it"... this seems the correct approach to take overall (using log4j), and I'm fine with just adding this, and adding it effectively disabled (ie trace leve), until we actually see some movement on the DataKind project.

Thanks!

omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
mseaton: It's not auto, I added this in just because Intellij gives an error, and this is it's suggested fix.

omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
mseaton: as above

omod/src/main/java/org/openmrs/module/pihcore/RequestMonitoringFilter.java
mseaton: It is available to log - it comes down to what you specify in your layout.  Any item in the ThreadContext can be included in the logging pattern / referenced by the particular appender as:  %X{key}

mseaton:
> I don't quite follow the format of the output, do you have an example?

So, for example if we wanted to mimic the current json format we are sending to data kind, we would use a pattern format like this:

```<PatternLayout pattern="\{&quot;timestamp&quot;: &quot;%d{ISO8601}&quot;, &quot;sessionId&quot;: &quot;%X{sessionId}&quot;, &quot;user&quot;: &quot;%X{user}&quot;, &quot;loadTime&quot;: &quot;%X{loadTime}&quot;, &quot;method&quot;: &quot;%X{method}&quot;, &quot;requestPath&quot;: &quot;%X{requestURI}&quot;, &quot;queryParams&quot;: &quot;%X{queryParams}&quot;\}%n" />```

**Comment by mseaton** [comment]
> I also think that we would need way to not have this stuff go to catalina.out, because I think it will make catalina.out bloated and hard to follow (or harder to follow than it currently is :)

Yeah, this is doable, I just needed to look it up.  You would do this by specifiying additivity="false" on the logger.

**Comment by mseaton** [comment]
> If the DataKind project does make headway, I think there will be a desire to have the new activity log format closely match the existing format, as they will likely have written some targeted scripts for slurping the logs.

So, this seems possible with the new logging.  We would first add a new appender to our log4j.xml file:

```
		<File name="ACTIVITY_LOG_APPENDER" fileName="${openmrs:applicationDirectory}/activity.log" append="true">
			<PatternLayout pattern="\{&quot;timestamp&quot;: &quot;%d{ISO8601}&quot;, &quot;sessionId&quot;: &quot;%X{sessionId}&quot;, &quot;user&quot;: &quot;%X{user}&quot;, &quot;loadTime&quot;: &quot;%X{loadTime}&quot;, &quot;method&quot;: &quot;%X{method}&quot;, &quot;requestPath&quot;: &quot;%X{requestURI}&quot;, &quot;queryParams&quot;: &quot;%X{queryParams}&quot;\}%n" />
		</File>
```

And then we would add a logger like this:

```
		<Logger name="org.openmrs.module.pihcore.RequestMonitoringFilter" additivity="false" level="TRACE">
			<AppenderRef ref="ACTIVITY_LOG_APPENDER" />
		</Logger>
```

PR #402, opened 2022-10-19
Uhm 6704 - Initial MFA support

mseaton:


api/src/main/java/org/openmrs/module/pihcore/metadata/Metadata.java
```diff
@@ -79,4 +88,44 @@ public static Concept getConcept(String lookup) {
         }
         return c;
     }
+
+    /**
+     * TODO: grant GET_LOCATIONS privilege to Anonymous instead of using a proxy privilege
+     */
```
mseaton, line 94: @mogoodrich do you think we should just go ahead and finally do this?  It seems right to me.  Then I can remove the proxy privilege stuff from here.

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
```diff
@@ -139,6 +140,7 @@ public void setupBase() throws Exception {
         removeOldPrivileges(); // TODO: This can likely be removed altogether at this point, or moved to liquibase
         ReportSetup.cleanupOldReports(); // TODO: could move this to liquibase or to a fixed scheduled task
         initializerMessageSource.addFallbackLanguage("ht", "fr");
+        AuthenticationConfig.registerClassLoader(getClass().getClassLoader());
```
mseaton, line 143: Because the authentication module is fully configurable via runtime properties in such a way that all of the classes are instantiated and configured by the authentication module itself, and not injected from other modules (eg. via Spring, or other patterns), I added the capability to register additional class loaders with the authentication module that it can use to find and create classes.  @mogoodrich I'd be interested in your thoughts and whether there are pitfalls to this approach or if you think it is generally reflective of a poor design.

omod/src/main/java/org/openmrs/module/pihcore/PihBasicAuthenticationScheme.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.http.HttpServletRequest;
+import java.util.List;
+import java.util.Properties;
+
+import static org.openmrs.module.appui.UiSessionContext.LOCATION_SESSION_ATTRIBUTE;
+import static org.openmrs.module.emr.EmrConstants.COOKIE_NAME_LAST_SESSION_LOCATION;
+
+/**
+ * This expands on the BasicWebAuthenticationScheme to also handle collecting, validating, and setting sessionLocation
+ */
```
mseaton, line 34: This class extends the Basic username/password authentication scheme provided by the authentication module, such that it also handles the selection of session location during initial basic authentication.

omod/src/main/java/org/openmrs/module/pihcore/PihBasicAuthenticationScheme.java
```diff
@@ -0,0 +1,89 @@
...
+	public void afterAuthenticationSuccess(AuthenticationSession session) {
+		super.afterAuthenticationSuccess(session);
+		Location loginLocation = getLoginLocation(session.getHttpRequest());
+		if (loginLocation != null) {
+			Context.getUserContext().setLocation(loginLocation);
+			session.setHttpSessionAttribute(LOCATION_SESSION_ATTRIBUTE, loginLocation.getLocationId());
+			session.setCookieValue(COOKIE_NAME_LAST_SESSION_LOCATION, loginLocation.getLocationId().toString());
+		}
+	}
+
```
mseaton, line 67: Because the OpenMRS AuthenticationScheme construct is not web-aware, and the authenticate method has no way to access the HTTP Session, I added some additional hooks into a WebAuthenticationScheme interface that allows for additional web-specific behavior to be included in an AuthenticationScheme that is executed before authentication, and after an authentication success or failure.  I'm using this to include the logic formerly in the LoginController post submission handler to retrieve the login location from the request, validate it, and set it in the various places (session, context, cookie).  NOTE:  This cookie is a _session_ cookie - meaning once the browser is closed it will be cleared.  If we actually intended this to be a persistent cookie with a longer TTL, we should set that, but AFAIK this is what we have had in place up until now

omod/src/main/webapp/pages/login.gsp
```diff
@@ -4,160 +4,166 @@
     ui.includeCss("pihcore", "login.css")
     ui.includeJavascript("pihcore", "account/login.js")
 %>
-
```
mseaton, line 7: You'll want to hide whitespace in the PR UI in order to see the changes in this file, I must have adjusted the indentations.

api/src/main/java/org/openmrs/module/pihcore/metadata/Metadata.java
mogoodrich: Seems reasonable?  I don't have a strong preference.

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
mseaton: Correct @mogoodrich .  The Authentication module uses an OpenMRS ModuleClassLoader instance, which has access to OpenMRS classes, and any module classes that it indicates are required modules or aware-of modules.  Since authentication is not aware of pihcore, it is not able to load a custom PIH authentication scheme by className.  So the solution I introduced was to allow downstream modules (eg. PIH core) to register classloaders with the authentication module, so that when the authentication module loads classes by name, it can use these classloaders to accomplish this.

An alternative would be to introduce a factory pattern (which is similar to the above, but abstracted away from the classloader level) and allow downstream modules (eg. PIH Core) to inject a custom factory to use to load specific classes.  For example, we could have an `AuthenticationSchemeFactory` interface in the authentication module, and then we could have an implementation of this in PIH Core, and then we could register this custom AuthenticationSchemeFactory via Spring autowiring or explicitly in the PIH Core Activator.

```
interface AuthenticationSchemeFactory {
    AuthenticationScheme newAuthenticationScheme(String schemeId);
}
```

This is possibly the cleaner approach?  Just requires an additional interface, and an additional custom class in each downstream module, and is less flexible - i.e. we'd need other factories for producing other types of custom classes - which is why I didn't go this route.

Happy to discuss and decide which to run with.

omod/src/main/java/org/openmrs/module/pihcore/PihBasicAuthenticationScheme.java
mseaton: Happy to discuss further @mogoodrich , but how would you aim to achieve this?  As I said, we have no easy way to do this in the authenticate(Credentials) method, which is what OpenMRS core supports.  We could add more information to the Credentials (eg. the location), but this doesn't really help because this still does not have access to the HttpSession for setting Location as a Session Attribute, or the HttpServletResponse for setting this Location as a Cookie, so we would still need these to be done outside of the authenticate method.  We could pass the entire AuthenticationSession (or just the Session and Response) in as part of the Credentials, but I am avoiding that because I think we need to keep the Credentials Serializable and light weight, and there are cyclical issues (Credentials are contained in the Session, so we can't really add the Session to the Credentials).

The solution above doesn't really "tack things on".  The alternative to the above (in design, but not in substance) would be to have a new authenticate method in a WebAuthenticationScheme base class which consumers would have to know to use in place of the existing one:

```
public Authenticated authenticate(Credentials credentials, AuthenticationSession session) {
   //  stuff to do before authentiation
   try {
       Authenticated authenticated = authenticate(credentials);
       // stuff to do after authentication succeeds
   }
   catch (Exception e) {
      // stuff to do after authentication fails
      if (!(e instanceof ContextAuthenticationException)) {
         e = new ContextAuthenticationException(e);
      }
      throw e; 
   }
}
```

If you have alternative or better ideas, let's please discuss.

omod/src/main/java/org/openmrs/module/pihcore/page/controller/LoginSecretPageController.java
```diff
@@ -0,0 +1,69 @@
...
+import org.openmrs.module.appui.UiSessionContext;
+import org.openmrs.module.authentication.AuthenticationContext;
+import org.openmrs.module.authentication.web.AuthenticationSession;
+import org.openmrs.module.pihcore.config.Config;
+import org.openmrs.ui.framework.UiUtils;
+import org.openmrs.ui.framework.annotation.SpringBean;
+import org.openmrs.ui.framework.page.PageModel;
+import org.openmrs.ui.framework.page.PageRequest;
+
+public class LoginSecretPageController {
```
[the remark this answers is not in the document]
mseaton: Yes, a particular AuthenticationScheme (eg. the TwoFactorAuthenticationScheme in the authentication module) would be configured (in runtime properties typically) to redirect to this page when secondary authentication is needed and this is the chosen option.

omod/src/main/java/org/openmrs/module/pihcore/PihBasicAuthenticationScheme.java
mogoodrich: Oops, sorry I wasn't clear in my comment. I was saying that your new approach was better than the *previous* tacked on approach... not suggesting *your* approach was tacked on...  sorry again, I should have just said "nice work" :)

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
mogoodrich: Happy to discuss further if you like, but also happy to run with whichever of the approaches you think it best... I just was trying to make sure I understood how it worked.

mogoodrich: Actually, maybe just add a comment here explaining what this is for?

mogoodrich (approved):
Added a minor comment, but LGTM

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
mseaton: done

PR #403, opened 2022-10-27
Allow session location to be null if authenticating via Authorization…

mseaton:
… headers, for backwards-compatibility


mogoodrich (approved):
LGTM!

PR #404, opened 2022-11-02
UHM-6804 - Enable users to administer their security question/answer

mseaton:
secret-question-admin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
 	</modules>
 
     <properties>
-        <openMRSVersion>2.5.7</openMRSVersion>
+        <openMRSVersion>2.5.8-SNAPSHOT</openMRSVersion>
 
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -46,7 +46,7 @@
 	</modules>
 
     <properties>
-        <openMRSVersion>2.5.7</openMRSVersion>
+        <openMRSVersion>2.5.8-SNAPSHOT</openMRSVersion>
```
mogoodrich, line 49: Probably okay, but was this intentional?

mogoodrich (approved):
Sweet, thanks @mseaton , LGTM, though see my one question....

pom.xml
mogoodrich: Ah, nevermind, just saw your comment regarding this on the ticket now.

PR #408, opened 2022-11-09
UHM-3909 - Months not translated for vaccination form

mseaton:


omod/src/main/webapp/pages/visit/templates/vaccination/vaccinationTable.gsp
```diff
@@ -33,7 +33,7 @@
                 <th>{{ vaccination.label | translate }}</th>
                 <td nowrap ng-repeat="sequence in sequences" ng-class="{ impossible: !isDoseValidForVaccination(sequence, vaccination) }">
                     <span ng-show="existingDose(sequence, vaccination)" ng-class="{'bold': existingDoseInEncounter(sequence, vaccination)}">
-                        {{ existingDose(sequence, vaccination) | groupMember:Concepts.vaccinationDate | obs:"value" | serverDate : 'dd-MMM-yyyy' }}
+                        {{ existingDose(sequence, vaccination) | groupMember:Concepts.vaccinationDate | obsValueDateLocalized }}
                         <a class="delete-action" ng-show="canDelete(sequence, vaccination)" ng-click="confirmDelete(sequence, vaccination)"><i class="icon-remove"></i></a>
```
mseaton, line 37: I could not figure out how to get the DatetimeFormats.dateLocalized constant from constants.js to be accessible from within the vaccinationTable gsp, but since the filters seem to be visible, and there is a pattern for using the constants in the filters that works, I just created a new filter to do what I needed to do, and then used the  filter here.

omod/src/main/webapp/pages/visit/visit.gsp
```diff
@@ -40,6 +40,10 @@
     ui.includeJavascript("pihcore", "visit/encounterTransaction.js")
     ui.includeJavascript("pihcore", "services/configService.js")
     ui.includeJavascript("file", "configuration/pih/scripts/visit/encounterTypeConfig.js", /* priority= */ null, /* pathIsRelativeToScripts= */ false)
+
+    if (locale != "en") {
+        ui.includeJavascript("pihcore", "angular-locale/angular-locale_" + locale + ".js")
+    }
 %>
```
mseaton, line 47: It seems that the only way to localize the angular-ui datepicker is to load in an appropriate locale file on the page, which sets various constant bindings that the datepicker uses.  I downloaded a french and spanish version of these and added them to the module, and then load these in if necessary.  I left it open-ended in case we need to add additional files for additional locales (presumably it will simply fail gracefully with a 404 if the JS file referenced here doesn't actually exist).

mseaton (commented):
@mogoodrich - I added some comments to explain the changes.

PR #412, opened 2022-11-21
UHM-6717 - Remove usage of privilege level full

mseaton:


api/src/main/resources/liquibase.xml
```diff
@@ -1185,5 +1185,25 @@
...
+                <sqlCheck expectedResult="0">
+                    select count(*) from user_role where role = 'Privilege Level: Full';
+                </sqlCheck>
+            </not>
+        </preConditions>
+        <comment>
+            Remove use of Privilege Level: Full in favor of Privilege Level: High for all users
+        </comment>
+        <sql>
+            create temporary table temp_priv_high_users as select user_id from user_role where role = 'Privilege Level: High';
```
mogoodrich, line 1200: This is just to prevent a duplicate row being created for users that somehow have both privileges?

mogoodrich (approved):
LGTM, thanks @mseaton

api/src/main/resources/liquibase.xml
mogoodrich: Makes sense.

PR #413, opened 2022-11-22
Move login pages

mseaton:


omod/src/main/webapp/pages/account/manageAccounts.gsp
```diff
@@ -45,7 +45,9 @@
 			</td>
 			<td>
 				<% if(it.username && it.username != '') {%>
-					${ ui.format(it.username) }
+					<a href="/${ contextPath }/authenticationui/account/userAccount.page?userId=${ it.user.id }">
+						${ ui.format(it.username) }
+					</a>
 				<% } %>
```
mseaton, line 51: This is the one place where we might need to discuss.  This provides a link into the user account page in the authenticationui module from the username column, if the account has a user.  The previous/legacy functionality (below) of editing using the big edit form still exists, but this link now allows one to get into the user's manage account page and get access to the various actions there (change 2fa, secret question, etc)

mogoodrich: If I'm understanding, I think this makes sense to me... ie, we support accessing both the PIH account management page and the new auth UI user admin stuff from this account list?  seems like a good compromise to trying to integrate those two pages.

mogoodrich (approved):
Sweet, thanks @mseaton ... and thanks for pointing out that key change (I agree with your approach, if I'm understanding it correctly).

PR #414, opened 2022-11-28
UHM-6876

mseaton:
I'm abandoning the authenticationui module for now, but while I was working on extracting the pages out of the PIH EMR, I found a few things that needed to be fixed.  That's what this PR is doing:

* Changing from using emr module constants to uicommons constants
* Moving the boillerplate template code out of all login pages into a decorator
* Limiting browser warning to just the initial basic authentication page

### DIFF
--- a/omod/src/main/webapp/resources/scripts/account/login.js
+++ b/omod/src/main/webapp/resources/scripts/account/login.js
@@ -1,3 +1,0 @@
-function isUsernameValid(username) {
-    return (username && username.length !== 0 && username.indexOf(' ') < 0);
-}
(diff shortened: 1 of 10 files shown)


omod/src/main/webapp/resources/scripts/account/login.js
```diff
@@ -1,3 +0,0 @@
-function isUsernameValid(username) {
-    return (username && username.length !== 0 && username.indexOf(' ') < 0);
```
mogoodrich, line 2: This is no longer needed? (Or maybe was just never used anymore?)

mogoodrich (approved):
Generally LGTM.  One small question on the PR.  Makes sense to extract shared functionality into a shared login page, but I didn't go through the gsp refactoring with a fine-toothed comb.

PR #418, opened 2023-01-12
add json file-types to authentication whitelist

mogoodrich:


mogoodrich:
@mseaton happy to defer on this if you think there are any risks I'm not thinking of.,,,

api/src/main/java/org/openmrs/module/pihcore/setup/AuthenticationSetup.java
```diff
@@ -58,6 +58,7 @@ public static void setup(Config config) {
         whitelist.add("*.ico");
         whitelist.add("*.ttf");
         whitelist.add("*.woff");
+        whitelist.add("*.json");
```
mseaton, line 61: Well, if we have an endpoints that accept ".json" and return protected data, this would bypass authentication checks outside of any service layer.

mseaton (approved):
I _think_ this should be ok, but not 100%.

api/src/main/java/org/openmrs/module/pihcore/setup/AuthenticationSetup.java
mogoodrich: fair, we probably want to think about this a bit more...

mogoodrich:
Going to close this out until  we are sure it is needed.

PR #419, opened 2023-01-13
UHM-6979 - Disable functionality to import patient from Lacolline

mseaton:


mogoodrich (approved):
Okay with me... as you would have guessed, part of me would have just commented all this out, but I'm okay with removal as we can dig up this commit if we need to recover anything.

PR #421, opened 2023-01-27
Uhm 6965

mseaton:


api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+        patientKeys.put("visit_id", "visit");
+        patientKeys.put("allergy_id", "allergy");
+        patientKeys.put("appointment_id", "appointmentscheduling_appointment");
+    }
+
+    @Override
+    public void startup() {
+        try {
+            statusDb = new Rocks(new File(config.getContext().getModuleDataDir(), "status.db"));
+            snapshotInitialized = BooleanUtils.isTrue(statusDb.get("snapshotInitialized"));
```
mseaton, line 60: Rocks DB is a key/value store, fast access, which I'm experimenting with for state management here.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+
+    @Override
+    public void shutdown() {
+        statusDb.close();
+    }
+
+    @Override
+    public void accept(DbEvent event) {
+        if (!snapshotInitialized) {
+            performInitialSnapshot();
```
mseaton, line 75: I recently moved the initialization code in here, as this ensures that it doesn't run until after the Debezium engine has started up and such.  I previously had it in the startup method.  We may want to put it back there (it fits there more naturally), I only moved it here as I was debugging some issues and thought this might be the cause.  The problem with putting it here is that it won't execute until the first event comes in.  So if the server is fired up in the middle of the night, this will be idle until someone (or a scheduled task) comes along and saves some sort of patient data, at which point this method will get triggered, it'll see that it needs to be initialized, and it'll do so, which will take some minutes to perform on a large DB given it's querying every patient table (including every Obs).

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+            statusDb.put("snapshotInitialized", Boolean.TRUE);
+            snapshotInitialized = true;
+        }
+        Set<Integer> patientIds = getPatientIdsForEvent(event);
```
mseaton, line 79: This is plural only really because of "relationship", which could theoretically refer to updates of 2 different patients.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    /**
+     * For the given event, returns the associated patient ids
+     * @param event the event to retrieve the related patient ids
+     * @return the patient ids associated with the given event
+     */
+    public Set<Integer> getPatientIdsForEvent(DbEvent event) {
+        Set<Integer> ret = new HashSet<>();
```
mseaton, line 96: The goal of this method is to get the patient_id associated with a given event.  This is easy when "patient_id" is a column on the table.  Otherwise, we need to determine how to retrieve the patient id for the given table and column(s).  This - as far as I've figured - covers all of our use cases.  I have other solutions I've played around with that are more generic.  We can evolve this as we see fit

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+        return ret;
+    }
+
+    /**
+     * Once a particular event is connected ot a particular patient, this tracks this change for the patient
+     * @param patientId the patient id related to the event
+     * @param event the event to track
+     */
+    public void handlePatientEvent(Integer patientId, DbEvent event) {
+        if (log.isDebugEnabled()) {
```
mseaton, line 133: For every event that we determine is an event for a particular patient, we update the db_event table with the timestamp of this event (based on the when the event occurred in the MySQL binlog), and also indicate whether the patient is deleted (based on either a hard delete of the row in the DB or based on a soft-delete if the patient is voided)

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Rather than stream all initial READ events one by one, it is much more efficient to set up the initial state
+     * by querying the database directly to record the most recent date updated and deleted status for all patients.
+     * This method performs this operation, and returns the max last_updated in the database
+     */
+    protected synchronized void performInitialSnapshot() {
+        if (!snapshotInitialized) {
+            log.warn("Performing initial snapshot into dbevent_patient from patient");
```
mseaton, line 158: In my initial implementation, I didn't do this.  I streamed the snapshot READ events in just like with the create/update/delete events, and if the event was a READ, I updated the table based on the date_created, date_changed, and date_voided columns.  But this did not perform well (took too long) on large existing DBs like HUM.  So instead, I wrote this method to perform a custom snapshot with our end-goal in mind, using DB optimizations to arrive at the same results.  On my system, this completes in 5-10 minutes on a HUM DB, which is perfectly acceptable.  We can discuss whether we even need any initialized values, depending on our downstream use case.  It seemed good to try to generate the best initial representation of when the most recent update for a patient likely took place, and then refine that based on streaming event data.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,209 @@
...
+                    List<String> dateCols = new ArrayList<>();
+                    if (columns.contains("date_created")) {
+                        dateCols.add("ifnull(t.date_created, p.last_updated)");
+                    }
+                    if (columns.contains("date_changed")) {
+                        dateCols.add("ifnull(t.date_changed, p.last_updated)");
+                    }
+                    if (columns.contains("date_voided")) {
+                        dateCols.add("ifnull(t.date_voided, p.last_updated)");
+                    }
```
mseaton, line 177: If we identify other columns that we want to factor in, we can add them here.  I wasn't immediately aware of any.

api/src/main/java/org/openmrs/module/pihcore/setup/DbEventSetup.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Setup debezium event sources and consumers
+     */
+    public static void setup() {
+        EventContext ctx = new EventContext();
+        DbEventSourceConfig config = new DbEventSourceConfig(100002,"PatientChangeSource", ctx);
+
+        // This configures no initial data snapshot to occur, as this is done as a part of the consumer startup
+        config.setProperty("snapshot.mode", "schema_only");
```
mseaton, line 24: This setting is what tells Debezium not to take an initial consistent snapshot of the data.  So we will not get millions of initial READ events sent to our consumer, which we don't need because we are performing our own custom snapshot.

api/src/main/java/org/openmrs/module/pihcore/setup/DbEventSetup.java
```diff
@@ -0,0 +1,34 @@
...
+
+        // Configure this source to monitor all patient-related tables
+        Set<String> patientTables = ctx.getDatabase().getMetadata().getPatientTableNames();
+        config.configureTablesToInclude(patientTables);
```
mseaton, line 28: See the dbevent module for how this is defined.

api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
```diff
@@ -42,89 +43,108 @@ public class PihCloseStaleVisitsTask implements Runnable {
 
     private static final String ENCOUNTER_TYPE_ED_TRIAGE_UUID = PihEmrConfigConstants.ENCOUNTERTYPE_EMERGENCY_TRIAGE_UUID;
 
+    private static boolean isExecuting = false;
+
     @Override
     public void run() {
```
mseaton, line 49: I can move all of this to a separate PR if desired.  Most of these changes are just whitespace.  I just added some additional logging and a check to ensure this doesn't run if it is already running.

api/src/test/java/org/openmrs/module/pihcore/event/PatientEventConsumerTest.java
```diff
@@ -0,0 +1,40 @@
+package org.openmrs.module.pihcore.event;
+
+import org.junit.jupiter.api.Test;
+
+public class PatientEventConsumerTest {
+
+    @Test
+    public void shouldTrackPatientChanges() throws Exception {
+        /*
```
mseaton, line 9: This was working when I wrote this in the dbevent module, I'll see if it can get it working in pihcore, but I may not want to do so.  The use of MySQL test containers is a bit complicated with running our builds in CI in docker itself...

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
```diff
@@ -67,6 +68,7 @@ public void configureSystem() {
             else {
                 configurationSetup.configureConceptDependencies();
             }
+            DbEventSetup.setup();
```
mogoodrich, line 71: can we make this configurable by a runtime property, at least in the short-term, so we can phase this in and start using it on only one or two CI servers?

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        else {
+            for (String key : patientKeys.keySet()) {
+                String keyTable = patientKeys.get(key);
+                Integer value = event.getValues().getInteger(key);
+                if (value != null) {
+                    SqlBuilder sql = new SqlBuilder();
+                    sql.select("p.patient_id").from("patient", "p");
+                    if (keyTable.equals("person")) {
+                        sql.where("p.patient_id = ?");
```
mogoodrich, line 108: Does't this not really do anything?  Or is this just future-proofing for the case when person_id != patient_id?

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,212 @@
...
+                    } else {
+                        sql.innerJoin(keyTable, "x", "patient_id", "p", "patient_id");
+                        sql.where("x." + key + " = ?");
+                    }
+                    patientId = database.executeQuery(sql.toString(), new ScalarHandler<>(1), value);
+                    if (patientId == null) {
+                        if (!keyTable.equals("person")) {
```
mogoodrich, line 115: Why is this okay in the person case?

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,212 @@
...
+        this.database = config.getContext().getDatabase();
+        patientKeys.put("person_id", "person");
+        patientKeys.put("person_a", "person");
+        patientKeys.put("person_b", "person");
+        patientKeys.put("order_id", "orders");
+        patientKeys.put("patient_program_id", "patient_program");
+        patientKeys.put("encounter_id", "encounter");
+        patientKeys.put("visit_id", "visit");
+        patientKeys.put("allergy_id", "allergy");
+        patientKeys.put("appointment_id", "appointmentscheduling_appointment");
```
mogoodrich, line 52: Do we need obs here?

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -0,0 +1,212 @@
...
+                String keyTable = patientKeys.get(key);
+                Integer value = event.getValues().getInteger(key);
+                if (value != null) {
+                    SqlBuilder sql = new SqlBuilder();
+                    sql.select("p.patient_id").from("patient", "p");
+                    if (keyTable.equals("person")) {
+                        sql.where("p.patient_id = ?");
+                    } else {
+                        sql.innerJoin(keyTable, "x", "patient_id", "p", "patient_id");
+                        sql.where("x." + key + " = ?");
```
mogoodrich, line 111: So, for example, there might be a change to the encounter table, but if the patient_id doesn't change (which likely never will!) it won't be in the event, so we fetch the patient id this way? Makes sense.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mogoodrich: Yeah, we need to think about performance (but agree that 5-10 minute seems perfectly reasonable)

mogoodrich (approved):
A few questions, but looks great. It's a pain, but can we get a conditional in place so we don't immediately enable this on all servers before merging?

api/src/main/java/org/openmrs/module/pihcore/PihCoreActivator.java
mseaton: Absolutely, good point.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mseaton: No.  I may rework the approach here.  This basically represents the possible joins based on key found in the event.   eg. if a table has "order_id", then it needs to join on "orders".  Obs has a person_id, so it's covered by the first entry.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mseaton: It's needed because there are persons in the DB that are not connected to patients.  So if a change to, say, person_name comes through for the mgoodrich user account, this will not result in an upsert to the dbevent_patient table since there is no associated patient record.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mseaton: Not exactly.  The patient_id is always in the event (the event contains the full before and after values, not just the diffs).  This piece of code wouldn't ever get hit for the "encounter" table, as an encounter event has a patient_id directly, so it would just get used.  This piece of code would get hit for something like "encounter_provider", which is patient data but only links to the patient via the encounter table.  So we need to first join encounter_provider to encounter in order to retrieve the associated patient_id from encounter.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mseaton: Because there will be situations where we are handling a DbEvent (since this is only filterable on the table level) for person-related tables which are not actually associated with patients (see my comment above).  So we don't throw an error in this case, but (assume) that this is a legitimate situation.

api/src/test/java/org/openmrs/module/pihcore/event/PatientEventConsumerTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@ExtendWith(MysqlExtension.class)
+public class PatientEventConsumerTest {
+
+    @Test
```
mseaton, line 25: Note, @mogoodrich I haven't gotten unit tests working for this yet, and we'll need to discuss the approach given our CI situation (eg. how to use Docker effectively both as an execution environment for the Maven build and also use testcontainers in our unit tests).

api/src/main/java/org/openmrs/module/pihcore/task/PihCloseStaleVisitsTask.java
mogoodrich: to clarify my previous comment, don't worry about moving to a separate PR unless you want to get this in faster.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mogoodrich: Ah right, got it.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mogoodrich: aha, that makes sense, and explains my confusion regarding obs as well, thanks....

mseaton:
@mogoodrich - new commits added to address some of your concerns above.  Can you have a quick look?

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
```diff
@@ -57,6 +57,11 @@ public void startup() {
         try {
             statusDb = new Rocks(new File(config.getContext().getModuleDataDir(), "status.db"));
             snapshotInitialized = BooleanUtils.isTrue(statusDb.get("snapshotInitialized"));
+            if (!snapshotInitialized) {
+                performInitialSnapshot();
+                statusDb.put("snapshotInitialized", Boolean.TRUE);
+                snapshotInitialized = true;
+            }
```
mogoodrich, line 64: I assume you determined it's okay to run this during startup now?

mogoodrich (approved):
LGTM... yeah, we can discuss getting some testing around this.

api/src/main/java/org/openmrs/module/pihcore/event/PatientUpdateEventConsumer.java
mseaton: Hard to know.  It will block startup until it completes, so that's the main factor - how long it takes on a system that hasn't run it before.  Maybe we should kick it off in a separate thread...

mogoodrich: Oh, right, I had forgotten the reasoning, I was thinking it was failing during startup... yeah, we can take a look at performance and decide, should keep it in mind the first time we roll out to a system with a large DB... a one-time 5 minute run seems fine to me, but we should remember not to panic the time that deployment starts... actually, thinking out probably worth putting a log.info or log.warn this to remind us if we are "log-watching" a prod startup.

mseaton: Ok, I just pushed a new commit that runs this setup in a separate thread and adds some additional logging as well

mogoodrich: Was trying to look at the new commit but Github appears a little flaky right now...the only issue with a separate thread, do we potentially have a risk of an issue if updates start coming in while it's initializing?  Also, do we need to makes sure this is initialized before any background tasks (like close stale visits?) runs?

mseaton: I don't think those things are issues.  But there is perhaps an issue with timing where there could be a small gap between the time when the initialization code runs and the binlog monitoring starts, during which time changes could happen that are not detected.  Let's discuss.

PR #422, opened 2023-01-30
have stale visits task only run once an hour

mogoodrich:
@mseaton  was reminded of this while looking at your changes to close stale visits and wanted to just get it in before I forgot


mseaton (approved):
Seems reasonable.

mogoodrich:
Thanks, merging!

PR #423, opened 2023-02-23
Add runtime/system property to explicitly disable the dbevent setup i…

mseaton:
…f desired.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
 	</modules>
 
     <properties>
-        <openMRSVersion>2.5.9-SNAPSHOT</openMRSVersion>
+        <openMRSVersion>2.5.11-SNAPSHOT</openMRSVersion>
 
         <addresshierarchyVersion>2.15.1</addresshierarchyVersion>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -46,7 +46,7 @@
 	</modules>
 
     <properties>
-        <openMRSVersion>2.5.9-SNAPSHOT</openMRSVersion>
+        <openMRSVersion>2.5.11-SNAPSHOT</openMRSVersion>
```
mogoodrich, line 49: Nice catch that this was behind... do we want/need the latest snapshot, or should we stick with 2.5.10 (I do see that Ian has added one bump fix to 2.5.11-SNAPSHOT)

mogoodrich (approved):
LGTM, one tangental question about the version of Core we want to be using.

pom.xml
mseaton: Not sure.  I also saw his fix, and figured we were on the snapshot before, probably best to stay on snapshots in the maintenance line.

PR #424, opened 2023-03-06
LIB-182: Registration: Add Homeless Checkbox

mogoodrich:
Note there were some tweaks to the registration app module and the ui commons module I needed to make to get this to work, see ticket:

https://pihemr.atlassian.net/browse/LIB-182

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsLiberia.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsLiberia.java
@@ -2,6 +2,8 @@ package org.openmrs.module.pihcore.apploader.apps.patientregistration;
 
 import org.openmrs.module.pihcore.config.Components;
 import org.openmrs.module.pihcore.config.Config;
+import org.openmrs.module.pihcore.metadata.Metadata;
+import org.openmrs.module.registrationapp.model.CheckboxWidget;
 import org.openmrs.module.registrationapp.model.DropdownWidget;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsLiberia.java
```diff
@@ -34,6 +36,33 @@ public void addSections(RegistrationAppConfig c) {
...
+        w.getConfig().setValue("true");
+        field.setWidget(toObjectNode(w));
+        homelessQuestion.addField(field);
+
+        // insert after the "address" question (assumption: address is first question)
+        contactInfoSection.getQuestions().add(1, homelessQuestion);
+
+        return contactInfoSection;
+    }
+
```
mseaton, line 65: This all looks fine, I guess, but FWIW, I do not think modeling this as a Person Attribute is a good idea.  Person Attributes are best used for things that are (ideally unchanging) attributes of a Person.  "Is Test Patient".  "Eye Color".  "Birthplace".  These things are set once, just like known attributes like Birthdate, and never (or rarely) changed.

Homeless is (hopefully) a temporary state that a person is in.  If they are homeless at the time of registration, we should record that as an Obs on their registration encounter.  If we want to see if this has changed over time, we can collect additional answers to this question in subsequent encounters.  This would seem a more appropriate way to model this.

@mogoodrich / @lnball ?

mseaton (commented):
If we keep this as an attribute, code itself looks fine @mogoodrich .  But I question whether that's the best way to model this, as I described.

PR #425, opened 2023-03-15
LIB-193 - Add CHW fields to Registration form

mseaton:


lnball (approved):
Looks good.  Assume there's also the labels added to pihemr/message properties?

PR #427, opened 2023-03-16
SL-293, replace ng-grid table with bootstrap pagination table

cioan:
@mogoodrich , I have replaced the ngGrid table with standard bootstrap table. I just had to include a newer version of bootstrap and some css that does the pagination. Here is how it looks now:

Screenshot 2023-03-16 at 6 35 11 PM


mogoodrich:
Thanks @cioan , I will check out tomorrow...

omod/src/main/webapp/pages/reports/activeVisitsList.gsp
```diff
@@ -1,6 +1,5 @@
 <%
-    ui.decorateWith("appui", "standardEmrPage")
-    ui.includeCss("pihcore", "inpatient.css")
+    ui.decorateWith("appui", "standardEmrPage", [ includeBootstrap: false ])
```
mogoodrich, line 2: Can we include a comment here saying that we are excluding the version of bootstrap provided by the UI Framework core in order to use a more recent version that we are providing (and list both versions) just as a reminder when we need to edit/debug this the future to make it clear which version of bootstrap we are using)(=

omod/src/main/webapp/pages/reports/activeVisitsList.gsp
```diff
@@ -26,12 +23,16 @@
...
-    ui.includeCss("uicommons", "angular-ui/ng-grid.min.css")
-
+    // ui.includeCss("uicommons", "angular-ui/ng-grid.min.css")
+    ui.includeCss("pihcore", "inpatient.css")
     ui.includeCss("pihcore", "activeVisitsList.css")
     ui.includeJavascript("pihcore", "visit/filters.js")
     ui.includeJavascript("pihcore", "reports/activeVisitsListController.js")
 
+    // libs for the bootstrap date picker and other libs
+//  the version in openmrs is older
```
mogoodrich, line 33: Oh re: my comment above, I see you've included the comment here I was talking about ... maybe just update this comment to note that we exclude the core version of bootstrap above and list the version we are using here?

PR #434, opened 2023-04-20
LIB-242 Removed patient support section for Liberia registration

lnball:
This is to remove the Patient support (CHW) questions from Liberia registration (ughhh) since it's now in check-in.


mogoodrich (approved):
LGTM...

I just disabled the Patient Support section in Liberia smoke test for now, but we likely should remove the code as well if it isn't used anywhere:

https://github.com/PIH/mirebalais-smoke-tests/commit/b03a24b64a49dcd09ccacf4380ef32e3dc47c341

PR #440, opened 2023-07-12
UHM-7260: Add Patient Flags module to the PIH EMR distro

mogoodrich:


mseaton (approved):
Code looks fine, if you think the module is in good enough shape to add in, in it's current form.

mogoodrich:
Maybe this is famous last words, but I don't think it can hurt anything in it's current shape, especially if we don't have any flags set up.  That being said, I think it may introduce some AOP, so maybe better to hold on this until after the LIberia release.

mseaton:
> I don't think it can hurt anything in it's current shape

Just add 5 new tables to the data model, add AOP around every EncounterService save, add additional spring context stuff that needs to startup, and add more modules to maintain, but yeah.

PR #441, opened 2023-07-12
UHM-7254, add dashboard widget to display recent lab results

cioan:
@mogoodrich , here is the PR for creating a new dashboard widget. Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
@@ -3204,22 +3204,19 @@ public class CustomAppLoaderFactory implements AppFrameworkFactory {
                 null));
 
         if (config.getCountry() == ConfigDescriptor.Country.SIERRA_LEONE) {
-            apps.add(addToClinicianDashboardFirstColumn(app(CustomAppLoaderConstants.Apps.ALL_LAB_RESULTS,
+            apps.add(addToClinicianDashboardFirstColumn(app(CustomAppLoaderConstants.Apps.RECENT_LAB_RESULTS,
                             "pihcore.lab.lab_results.caps",
                             "fas fa-vial",
                             null,
                             null,
                             objectNode(
-                                    "widget", "latestObsForConceptList",
-                                    "icon", "fas fa-vial",
-                                    "label", "pihcore.lab.lab_results.caps",
-                                    "concepts",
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -3204,22 +3204,19 @@ private void enableLabs() throws UnsupportedEncodingException {
...
                             null,
                             null,
                             objectNode(
-                                    "widget", "latestObsForConceptList",
-                                    "icon", "fas fa-vial",
-                                    "label", "pihcore.lab.lab_results.caps",
-                                    "concepts",
-                                        CustomAppLoaderConstants.SARS_COV2_ANTIBODY_TEST + "," + CustomAppLoaderConstants.SARS_COV2_ANTIGEN_TEST + "," + CustomAppLoaderConstants.SARS_COV2_RT_PCR_TEST + "," + CustomAppLoaderConstants.SARS_COV2_XPERT_TEST + "," + CustomAppLoaderConstants.GLUCOSE_TEST + "," + CustomAppLoaderConstants.BUN_TEST + "," + CustomAppLoaderConstants.NA_TEST + "," + CustomAppLoaderConstants.K_TEST + "," + CustomAppLoaderConstants.CR_TEST + "," + CustomAppLoaderConstants.HB_TEST,
-                                    "conceptNameType", "shortName",
+                                    "encounterTypes", PihEmrConfigConstants.ENCOUNTERTYPE_LAB_SPECIMEN_COLLECTION_UUID + "," + PihEmrConfigConstants.ENCOUNTERTYPE_LAB_RESULTS_UUID,
```
mogoodrich, line 3213: It's definitely confusing how we do all this, but I think we want to pull the encounter types from this global property, which is what I believe the Lab Workflow OWA uses...

https://github.com/PIH/openmrs-config-pihemr/blob/master/configuration/globalproperties/gp.xml#L327

omod/src/main/webapp/fragments/dashboardwidgets/labResults.gsp
```diff
@@ -0,0 +1,28 @@
...
+            <li>
+                <div style="display: block; overflow: hidden; padding-right: 5px; padding-bottom: 2px">
+                    <span style="float: left">
+                        ${ui.format(labResult.concept.getShortNameInLocale(ui.locale) != null ? labResult.concept.getShortNameInLocale(ui.locale) : labResult.concept.getPreferredName(ui.locale))}:&nbsp;
+                    </span>
+                    <span style="float: left">
+                        <strong>${ ui.format(labResult) }</strong>
+                    </span>
+                    <span style="float: right; font-size: small; color: #939393">
+                        ${ ui.formatDateWithClientTimezone(labResult.obsDatetime) }
```
mogoodrich, line 21: I believe we just want to use formatDatePretty here.  Are there other cases when we use the "withClientTimezone" functionality directly?

In the formatDatePretty, method you will see that it only uses the client time zone if the time zone version GP is set to true (which we have as false in our implementation)

I hate timezomes, but the explanation is that in our servers, the server timezone always maps to the clinic timezone, and that is the date we want to display.  ie, if the lab result came in on May 21, 2023 in the clinic timezone, that's the date we want to display, not the day before if per chance the client is in a different timezone.  All this "clientTimezone" stuff was added by Mekom because they generally keep their servers in UTC, regardless of the location of the clinic.

mogoodrich (changes requested):
Thanks @cioan !  Looks great, with a note of a couple changes to make.

cioan:
Thanks @mogoodrich ! I just fixed the things you pointed above and committed those changes. Thanks!

PR #442, opened 2023-08-10
SL-366 Set lab order duration from ncd htmlform

mseaton:


mogoodrich (approved):
Generally looks good to me, maybe some test cases if it's relatively straightforward to do?

PR #443, opened 2023-08-11
UHM-7072 - New Dispensed Meds Dashboard widget

mseaton:


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/MedsDispensedFragmentController.java
```diff
@@ -0,0 +1,172 @@
...
+
+        // Get all Med Dispensed records from the Domain
+        MedicationDispenseCriteria criteria = new MedicationDispenseCriteria();
+        criteria.setPatient(patient);
+        criteria.setIncludeVoided(false);
+        List<MedicationDispense> dispenses = medicationDispenseService.getMedicationDispenseByCriteria(criteria);
+        for (MedicationDispense d : dispenses) {
+            Concept medConcept = (d.getDrug() != null ? d.getDrug().getConcept() : d.getConcept());
+            Date dispenseDate = d.getDateHandedOver();
+            if (dispenseDate == null) {
```
mogoodrich, line 122: I need to check the workflow again, but it's possible to have a dispense row for a drug that is "on hold" or for a drug that is "prepared" but not handed over (or that is long-term intent) so I believe that we only want to include dispenses with a "date_handed_over" (assuming the business logic is only to display those drugs the patient has actually received).  

You should be able to look at the new Dispense app to confirm how it works (or I can remind myself later)

mogoodrich: Oh, I see, you are only checking the encounter... I don't think we associate dispensing with encounters, so this may be fine.

mseaton: OK, I'll change.  I didn't know off-hand what fields the dispensing esm was populating and when.  If I can always rely upon the "dateHandedOver" being populated when a medication has actually been dispensed, I'll just stick with that, and if it is null I won't include it.  Make sense @mogoodrich ?

mogoodrich (approved):
LGTM, thanks @mseaton  and sorry for the delay.

PR #446, opened 2023-09-21
SL-414, display drug orders concepts in the encounter header

cioan:
This is for displaying medication orders concept names in the encounter header:
Screenshot 2023-09-21 at 10 25 08 AM


cioan:
Thanks @mogoodrich !

PR #450, opened 2023-09-27
Creating another branch to prevent code break in branch HAI-753

Dutervil:
@cioan have a look on it. it is not showing in the alert.


cioan:
This looks clean, @Dutervil ! This looks like my original PR? Have you added any code to it?

Dutervil:
@mogoodrich have a look on it. it is not showing in the alert.

cioan (approved):
Hi @Dutervil , I am glad to hear that this works for you now. Here, i would just add a check for null. For example if the HIV Program Location is null. Then, if I remember the requirement correctly you have to check for the location of the Registration encounter? But, you could add  this later, if you want first to just push this code change and test it.

PR #451, opened 2023-09-29
UHM-7470, remove dossier column if identifier not defined

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

omod/src/main/webapp/pages/reports/inpatientList.gsp
```diff
@@ -80,8 +82,10 @@
     %>
     <tr id="visit-${ v.patientId
     }">
-        <td>${ v.zlEmrId ?: ''}</td>
-        <td>${ v.dossierNumber ?: ''}</td>
+        <td>${ v.zlEmrId ?: v.primaryIdentifier }</td>
```
mseaton, line 85: I'm pretty sure the zlemrid _is_ the primary identifier in Haiti, right?  Or is this sometimes different?  If it is the same, can't we just use primaryIdentifier everywhere?

mseaton (approved):
LGTM, though see my comment (I'm fine just merging this as-is though)

omod/src/main/webapp/pages/reports/inpatientList.gsp
cioan: You are correct, @mseaton ! I just replaced the zlEmrId with the primaryIdentifier. Thanks!

PR #452, opened 2023-10-07
UHM-7478, in Sierra Leone display template should include country

cioan:
The default address section was excluding the country field when displaying the address. So, I had to overwrite this method in the SierraLeone class.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
@@ -115,23 +115,6 @@ public class SectionsSierraLeone extends SectionsDefault {
 
         return q;
     }
-    /**
-     * In the base class this method excludes the Country field.
-     * In Sierra Leone we want to display the Country because we are capturing addresses from two other countries, Liberia and Guinea
-     * @param levels
-     * @return a String representing the full address
-     */
-    @Override
-    protected String getAddressHierarchyDisplayTemplate(List<AddressHierarchyLevel> levels) {
-        StringBuilder displayTemplate = new StringBuilder();
-        displayTemplate.append("{{nvl field.[" + levels.size() + "] '-'}}");
-        for (int i = levels.size() - 1; i >= 1; i--) {
-            displayTemplate.append(", {{field.[" + i + "]}}");
-        }
-
-        return displayTemplate.toString();
-    }
-
     private Question getLocalAddressQuestion() {
         Question q = new Question();
         q.setId("localAddressLabel");
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
@@ -115,6 +115,39 @@ public class SectionsSierraLeone extends SectionsDefault {
 
         return q;
     }
+
+    /**
+     * By default, this would generate a display template that removes the Country. In Sierra Leone,
+     * we are capturing addresses from 3 countries, Sierra Leone, Liberia and Guinea, therefore we
+     * need the country displayed
+     * @return Question
+     */
+    @Override
+    public Question getAddressQuestion() {
+        Question q = new Question();
+        q.setId("personAddressQuestion");
+        q.setLegend("registrationapp.patient.address");
+        q.setHeader("registrationapp.patient.address.question");
+
+        Field f = new Field();
+        f.setType("personAddress");
+
+        // If there are address hierarchy levels configured, use the address hierarchy widget, otherwise use the standard address widget
+        List<AddressHierarchyLevel> levels = Context.getService(AddressHierarchyService.class).getAddressHierarchyLevels();
+        if (levels != null && levels.size() > 0) {
+            f.setWidget(getAddressHierarchyWidget(levels, null, true));
+        }
+        else {
+            Map<String, String> m = new HashMap<String, String>();
+            m.put("providerName", "uicommons");
+            m.put("fragmentId", "field/personAddress");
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
```diff
@@ -115,6 +115,39 @@ private Question getLocalContactPhoneNumber() {
...
+        q.setLegend("registrationapp.patient.address");
+        q.setHeader("registrationapp.patient.address.question");
+
+        Field f = new Field();
+        f.setType("personAddress");
+
+        // If there are address hierarchy levels configured, use the address hierarchy widget, otherwise use the standard address widget
+        List<AddressHierarchyLevel> levels = Context.getService(AddressHierarchyService.class).getAddressHierarchyLevels();
+        if (levels != null && levels.size() > 0) {
+            f.setWidget(getAddressHierarchyWidget(levels, null, true));
```
mseaton, line 138: I'm not really sure I follow.  I had to look at the default code and compare to see what you actually changed here.  Would be good to be more specific about how this is different in the comments, or even better, call the super() method and then modify the returned question.
That said, it seems like the change you made is to remove the line above this one that reads:
`q.setDisplayTemplate(getAddressHierarchyDisplayTemplate(levels));`

Isn't the right solution to change the implementation of "getAddressHierarchyDisplayTemplate" to include country?

cioan:
Thanks @mseaton for reviewing this PR. I have just updated the PR with another commit. Yes, it was much easier to just overwrite the "getAddressHierarchyDisplayTemplate()" in the SierraLeone class to include the Country field. Thanks!

api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
```diff
@@ -115,6 +115,23 @@ private Question getLocalContactPhoneNumber() {
...
+    protected String getAddressHierarchyDisplayTemplate(List<AddressHierarchyLevel> levels) {
+        StringBuilder displayTemplate = new StringBuilder();
+        displayTemplate.append("{{nvl field.[" + levels.size() + "] '-'}}");
+        for (int i = levels.size() - 1; i >= 1; i--) {
+            displayTemplate.append(", {{field.[" + i + "]}}");
+        }
+
+        return displayTemplate.toString();
+    }
+
```
mseaton, line 134: This seems fine.

If we really want to try to improve on this overall and for the future, might opt instead to just modify the original method to support both use cases.  For example, this method could take in a boolean `showCountry` parameter, and if `true` then iterate down to level 1, and if `false` then iterate down to level 2.

Then, the `getLocalAddressQuestion` and `getAddressQuestion` methods could pull this `showCountry` configuration from the registration config or (even better) determine what to do based on whether more than one country are included in the address hierarchy configuration.

cioan: Thanks @mseaton ! I have updated this PR with your suggestion. I also opened a SL config with the change:
https://github.com/PIH/openmrs-config-pihsl/pull/82

api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
```diff
@@ -128,7 +128,7 @@ private Question getLocalAddressQuestion() {
         // If there are address hierarchy levels configured, use the address hierarchy widget, otherwise use the standard address widget
         List<AddressHierarchyLevel> levels = Context.getService(AddressHierarchyService.class).getAddressHierarchyLevels();
         if (levels != null && levels.size() > 0) {
-            //q.setDisplayTemplate(getAddressHierarchyDisplayTemplate(levels));
+            q.setDisplayTemplate(getAddressHierarchyDisplayTemplate(levels));
             f.setWidget(getAddressHierarchyWidget(levels, getLocalContactAddressFieldMappings(), true));
```
mseaton, line 132: Now that the result of `getAddressHierarchyDisplayTemplate(levels)` is able to be customized based on `isShowCountryInAddressHiercharchy()`, do we need this custom implementation of this method for Sierra Leone?  Is there a reason we can't uncomment this for everything?  Do we know why this was commented out in the first place and can we test that this works as expected (look at the git blame / git history and see what tickets are associated with this)?

mseaton (commented):
Looks good @cioan .  See my question on the SectionsSierraLeone class.

api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
cioan: Sorry @mseaton, I am not sure I understand the question here. The q.setDisplayTemplate(getAddressHierarchyDisplayTemplate(levels)) line has already been uncommented for Sierra Leone. 
Are you asking to uncomment this line in the SectionsHaiti.java class? That's the only other place I found this line to be commented out. I think it is very hard to know the history of this line. It looks like this was part of a massive refactoring effort when you moved code from the mirebalais module into the pihcore?

Screenshot 2023-10-10 at 12 21 35 AM

mseaton: Sorry, I misread what was happening here.  This isn't the main address question, but a custom "local address" question that you added back in June.  I assume you just copied this from the Haiti version, with the commented out line in it?  Makes sense to uncomment it now like you have done, so I think this is good. But if we copied this from somewhere (eg. Haiti) where this line is commented out, might be worth revisiting why it is commented out there and whether that is still valid given the improvements you've made here.

mseaton (approved):

cioan:
Thanks @mseaton !

PR #453, opened 2023-10-11
UHM-7497 - Add Locations for MCOE

mseaton:


api/src/main/java/org/openmrs/module/pihcore/config/Config.java
```diff
@@ -128,6 +128,18 @@ public Locale getIdCardLocale() {
...
+    public boolean isSierraLeone() {
+        return getCountry().equals(ConfigDescriptor.Country.SIERRA_LEONE);
+    }
+
+    public boolean isWellbody() {
+        return isSierraLeone() && getSite() != null && getSite().toUpperCase().startsWith("WELLBODY");
+    }
+
+    public boolean isKgh() {
+        return isSierraLeone() && getSite() != null && getSite().toUpperCase().startsWith("KGH");
```
mogoodrich, line 140: super nit, maybe a comment here to make remind us that that we specifically have the starts with because of KGH-Test

PR #454, opened 2023-10-18
UHM-7475, configure Sierra Leone with the new relationships widget

cioan:


api/src/main/java/org/openmrs/module/pihcore/config/registration/PersonRelationshipConfigDescriptor.java
```diff
@@ -0,0 +1,72 @@
+package org.openmrs.module.pihcore.config.registration;
+
+import org.codehaus.jackson.annotate.JsonProperty;
+
```
mseaton, line 4: I don't really understand this, when you already have a PersonRelationshipWidget.Config class in the registrationapp module that is essentially the same.  I would think we should convert the PersonRelationshipWidget.Config from a static inner class into a normal class that matches what you have here (essentially just adds the label property, right)?  And then just use that.  No need for there to be 2 of the exact same config class, is there?  Or am I missing how this should work?

mseaton (changes requested):

api/src/main/java/org/openmrs/module/pihcore/config/registration/PersonRelationshipConfigDescriptor.java
mogoodrich: Looking quickly at the existing code, I think this generally follows the existing pattern we've used for the other widgets, so it makes sense to continue like this.

The Config descriptor in Registration App contains the generic ways to configure the widget, and here we have the specific things we want allowable to be configured in the PIH EMR, which happens to be pretty much the same in this case?

mogoodrich (commented):
See my comment...

PR #455, opened 2023-10-23
Hai 753 update

Dutervil:
Hi @dmdesimone . Please check this


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/hiv/ProgramLocationWarningFragmentController.java
```diff
@@ -33,7 +33,12 @@ public void controller(FragmentModel model,
             if (patient !=null && hivProgram !=null) {
                 List<PatientProgram> patientPrograms = programWorkflowService.getPatientPrograms(patient, hivProgram, null, null, null, null, false);
                 if (patientPrograms != null && patientPrograms.size() > 0 ) {
-                    model.addAttribute("hivProgramLocation", patientPrograms.get(0).getLocation());
+                    if(patientPrograms.get(0).getLocation() !=null){
+                      model.addAttribute("hivProgramLocation", patientPrograms.get(0).getLocation());
+                    }else{
+                     model.addAttribute("error", "No Location Found for this patient");
```
cioan, line 39: Hi @Dutervil , this looks OK. Another way to safeguard agains program location null is to set the attribute up to the top:
` model.addAttribute("sessionLocation", uiSessionContext.getSessionLocation());
+        model.addAttribute("hivProgramLocation", null);`
and then in the gsp page you check if the location is null and write the same error.
It's basically the same thing but avoids the need to introduce another model parameter, 'error'. 
Please let me know if this makes sense.

omod/src/main/webapp/fragments/hiv/programLocationWarning.gsp
```diff
@@ -1,15 +1,18 @@
-
+<% hivProgramLocation.name !=null {%>
 <div>
 <div id="hiv-program-location-info" style="background-color: lightcoral; padding: 10px; border-radius: 5px;">
       <ul>
           <li>Session Location: <b>${sessionLocation.name}</b></li>
           <li>
+             <% if(error !=null){ %>
```
cioan, line 7: here you would check if hivProgramLocation != null and write the error message if necessary

cioan (changes requested):
I made a couple small suggestions.

Dutervil:
I think everything is good now @cioan

cioan:
Thanks @Dutervil ! Are you able to merge this PR? Or, do you want me to merge it in?

PR #460, opened 2023-11-27
UHM-7577 - Administer account should point to admin url

mseaton:


mogoodrich (approved):
@mseaton  not sure I totally follow, but if you and @dmdesimone  are okay with the workflow, it works for me... I know there was some confusion before about how to access the different pages to modify an account

PR #465, opened 2023-12-12
UHM-7589, add wristband template for Sierra Leone EMR

cioan:
@mogoodrich , I just created a separate template for implementing the requirements for the MCOE printing. For now I just copied the existing template from Mirebalais and deleted the fields that are not required at MCOE. Now, the Print Wristband action is visible in KGH Test instance and clicking print won't cause a stacktrace. I am going to continue to work on implementing the exact fields required at MCOE, but first just wanted to make sure this is the right approach? Thanks!

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/WristbandFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/WristbandFragmentController.java
@@ -4,7 +4,9 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openmrs.Location;
 import org.openmrs.Patient;
+import org.openmrs.api.context.Context;
 import org.openmrs.module.appui.UiSessionContext;
+import org.openmrs.module.pihcore.printer.template.SLWristbandTemplate;
 import org.openmrs.module.pihcore.printer.template.WristbandTemplate;
 import org.openmrs.module.paperrecord.PaperRecordService;
 import org.openmrs.module.paperrecord.UnableToPrintLabelException;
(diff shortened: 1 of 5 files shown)


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/WristbandFragmentController.java
```diff
@@ -33,19 +37,26 @@ public SimpleObject printWristband(UiUtils ui,
...
             }
 
-            // make sure a paper record has been created
-            if (!paperRecordService.paperRecordExistsForPatient(patient, location)) {
-                paperRecordService.createPaperRecord(patient, location);
+            String data = "";
+            if (config.getCountry().equals(Country.SIERRA_LEONE)) {
+                SLWristbandTemplate slWristbandTemplate = Context.getRegisteredComponent("slWristbandTemplate", SLWristbandTemplate.class);
+                data = slWristbandTemplate.generateWristband(patient, location, uiSessionContext.getLocale());
+            } else {
```
mogoodrich, line 52: Now that we are printing wristbands for multiple countires, I'd make ZL explicit as well, ie rename WristbandTemplate to ZLWristbandTemplate, get rid of the autowiring of the WristbandTemplate in the code above, and just fetch the "zlWristbandTemplate" if the country = HAITI, the same way we do for Sierra Leone

mogoodrich (approved):
LGTM, thanks @cioan , I added one comment.

The other thing I'd add would be that if the two wristbands are similar enough you could try using a single WristbandTemplate that takes in a config.getCountry() and customize within the teplate based on that... but I suspect you've already discovered that they are going to be fairly different?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/WristbandFragmentController.java
cioan: Thanks @mogoodrich ! I have refactored the ZlWristbandTemplate.

cioan:
Thanks @mogoodrich ! That's a good idea. As I start implementing the wristband fields for Sierra Leone, I will keep this in mind, and if it makes sense I will refactor the template. I think the bigger difference would be with the location of those fields on the wristband. The ZPL language and special characters makes it difficult to follow where each field should be located.

mogoodrich:
Yeah, I suspect it will be different enough that we will want to keep two files, but worth keeping in mind.  Agree that it's the potentially different layout more than the fields themselves.

PR #466, opened 2023-12-13
SL-462-2: New National ID field on Registration

mogoodrich:


mogoodrich:
Update so that the one-question-per-screen validator does regex validation using the regex validator assigned on the identifier type.

api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsSierraLeone.java
```diff
@@ -214,7 +217,8 @@ private Question getNationalId() {
         f.setFormFieldName("patientIdentifier" + SierraLeoneConfigConstants.PATIENTIDENTIFIERTYPE_NATIONALID_UUID);
         f.setUuid(SierraLeoneConfigConstants.PATIENTIDENTIFIERTYPE_NATIONALID_UUID);
         f.setType("patientIdentifier");
-        f.setWidget(getTextFieldWidget(16));
```
mseaton, line 217: Looks fine, though it would be a bit nicer if this was conditional on whether the nationalId.format was non-empty, as this would allow us to more easily just adjust in the config if needed, without needing a new gold to go along with it, if they ever decide that this regex is too strict and want to loosen it.

mseaton (approved):
Approving, but see my comment

PR #470, opened 2024-01-23
UHM-7682: MFA not working on HIV EMR (production)

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/setup/AuthenticationSetup.java
```diff
@@ -135,20 +135,18 @@ public static void setup(Config config) {
     }
 
     /**
-     * Add configuration for a scheme with the given schemeId, if a scheme with this schemeId is not already configured
+     * Add configuration for a scheme with the given schemeId
```
mogoodrich, line 138: So, basically, if the scheme was already configured (which would be the case when the "Refresh Entire Configuration" button is clicked, the block below would be skipped, and therefore the "loginPage" would not be added to the list of files to whitelist, causing the error when a user with MFA enabled tried to log in.

We could just change this so that the only change is to *always* whitelist the whitelist set, but it seemed like there is no reason not to just re-add the scheme even if it already exists, because it should always overwrite? I tested and it seened to work, thoughts @mseaton ?

mseaton: Yeah, I see no reason this shouldn't work for our usage.  The other approach we could take instead, would be to leave this as-is, but at the top of this method first reset the AuthenticationConfig. eg. somewhere around line 38 add:

```AuthenticationConfig.setConfig(new Properties());```

This would ensure that every time we reload the configuration, we are reloading the authentication config from scratch and the following methods would work consistently each time.

mseaton (approved):
I _think_ this should be fine, but also seem my alternative suggestion to consider.

api/src/main/java/org/openmrs/module/pihcore/setup/AuthenticationSetup.java
mogoodrich: yeah, I can look into that, that's probably a better approach to make sure we clear everything out.

PR #471, opened 2024-01-24
SL-469 Added checkin forms for mcoe and SL only

lnball:
Can you please check these updates?  It doesn't work perfectly yet.  It should do the following:

- For SL, there should be 4 checkin files: livecheckin, checkin, checkinMaternal, liveCheckinMaternal
- For Non-SL sites, there should be no changes
- For SL, the Maternal checkin forms will only appear for the MCOE Triage location (tag: Check In Maternal Location).  The regular and current checkin forms should appear for all other configured locations (tag: Check In Location).
- Unclear about the circular form from the home page app.  I have that code commented out in the CALF.

Questions:
- Can we use the same encounter type for the live/retro versions?
- Can we use the same actions for the live/retro versions (CHECK_IN_VISIT ACTION)?

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
@@ -591,25 +591,77 @@ public class CustomAppLoaderFactory implements AppFrameworkFactory {
                     //     "/registrationapp/registrationSummary.page?patientId={{patientId}}&breadcrumbOverrideProvider=coreapps&breadcrumbOverridePage=findpatient%2FfindPatient&breadcrumbOverrideApp=" + Apps.CHECK_IN + "&breadcrumbOverrideLabel=mirebalais.app.patientRegistration.checkin.label",
                     null, config.getFindPatientColumnConfig()),
                     sessionLocationHasTag("Check-In Location")));
+
+            /*
+            if (config.getCountry().equals(ConfigDescriptor.Country.SIERRA_LEONE) ) {
+                apps.add(addToHomePage(findPatientTemplateApp(CustomAppLoaderConstants.Apps.CHECK_IN,
+                                "mirebalais.app.patientRegistration.checkin.label",
+                                "fas fa-fw fa-paste",
+                                "App: mirebalais.checkin",
+                                "/pihcore/checkin/checkin.page?patientId={{patientId}}",
+                                //     "/registrationapp/registrationSummary.page?patientId={{patientId}}&breadcrumbOverrideProvider=coreapps&breadcrumbOverridePage=findpatient%2FfindPatient&breadcrumbOverrideApp=" + Apps.CHECK_IN + "&breadcrumbOverrideLabel=mirebalais.app.patientRegistration.checkin.label",
+                                null, config.getFindPatientColumnConfig()),
+                        sessionLocationHasTag("Check-In Maternal Location")));
+            } */
         }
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -591,25 +591,77 @@ private void enableCheckIn(Config config) {
...
+            extensions.add(overallRegistrationAction(CustomAppLoaderConstants.Extensions.CHECK_IN_REGISTRATION_ACTION,
+                    "mirebalais.maternalCheckin.label",
+                    "fas fa-fw icon-check-in",
+                    "link",
+                    enterSimpleHtmlFormLink(PihCoreUtil.getFormResource("liveCheckinMaternal.xml")) + andCreateVisit(),
+                    "Task: mirebalais.checkinForm",
+                    sessionLocationHasTag("Check-In Maternal Location")));
+        } else {
+            // check-in form that appears on visit and clinicial dashboard after a visit has been started as a "Visit Action"
+            extensions.add(visitAction(CustomAppLoaderConstants.Extensions.CHECK_IN_VISIT_ACTION,
```
mogoodrich, line 647: instead of modifying the CHECK_IN component, I would think it would just make sense to add the new MCOE check-in form to the existin MCOE_Forms component, and leave the check-in alone?

Also, not sure if we need both a "regular" and a "live" form?  I think we developed both at Mirebalais as a hack-ish way to hand the fact that they wanted a check-in form that never allowed editing date/provider and location.  to keep it simple I would start with the single "checkinMaternal.xm" until someone complains and wants a separate "live" one.

mogoodrich (changes requested):
See my comments/thoughts @lnball

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -591,25 +591,77 @@ private void enableCheckIn(Config config) {
...
+                    sessionLocationHasTag("Check-In Location")));
+
+            // Maternal check-in form that appears on visit and clinicial dashboard after a visit has been started as a "Visit Action"
+            extensions.add(visitAction(CustomAppLoaderConstants.Extensions.CHECK_IN_MATERNAL_VISIT_ACTION,
+                    "mirebalais.maternalCheckin.label",
+                    "fas fa-fw icon-check-in",
+                    "link",
+                    enterSimpleHtmlFormLink(PihCoreUtil.getFormResource("checkinMaternal.xml")),
+                    "Task: mirebalais.checkinForm",
+                    sessionLocationHasTag("Check-In Maternal Location")));
```
mogoodrich, line 634: We also might be able to just use the "Maternal and Child Location" we use for the other MCOE forms here?  Or just the basic Check-In tag?  Might make sense to just start with one of these two until we confirm that we need a custom location set for this form?

lnball: Right @mogoodrich .  I think we need a separate Check-In Maternal Location to handle the 2 checkin locations with different checkin encounter/forms for MCOE.  I'm fine with just the live form at MCOE (and not the live + retro).  I can simplify the code and you can review again?  Or if you can muster to energy to talk to me?

mogoodrich: That all sounds correct to me @lnball !

PR #473, opened 2024-01-29
SL-480: Add patient to queue via post submission action

cioan:
@mseaton and @mogoodrich , here is a simple PostSubmissionAction that adds a patient to a patient queue. I have modified the checkin form with the coded answer to point to the MCOE Triage service:

`<div class="section-container" style="color:black">
                <obs conceptId="CIEL:1272" style="checkbox"
                     answerConceptId="PIH:14976" answerCode="Refer to MCOE triage" />
     </div>
`
I thought that this way this action could be more generic, and potentially we could use it to add patient to other queues by just changing the html form.


api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openmrs.module.htmlformentry.FormEntrySession;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.module.queue.api.QueueServicesWrapper;
+import org.openmrs.module.queue.api.search.QueueSearchCriteria;
+import org.openmrs.module.queue.model.Queue;
+import org.openmrs.module.queue.model.QueueEntry;
+
+public class AddPatientToQueue implements org.openmrs.module.htmlformentry.CustomFormSubmissionAction {
```
mogoodrich, line 20: For consistency with the other classes, let's call this "AddPatientToQueueAction"

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                Concept serviceQueue = candidate.getValueCoded();
+                QueueSearchCriteria criteria = new QueueSearchCriteria();
+                criteria.setServices(Arrays.asList(serviceQueue));
+                QueueServicesWrapper queueServicesWrapper = Context.getRegisteredComponents(QueueServicesWrapper.class).get(0);
+                List<Queue> queues = queueServicesWrapper.getQueueService().getQueues(criteria);
+                Queue queue = null ;
+                QueueEntry queueEntry = null;
+                if ( !queues.isEmpty() ) {
+                    queue = queues.get(0);
+                }
```
mogoodrich, line 49: Let's log an error if no matching queue is found, and also log an error if multiple queues are found.  I assume the queue module allows you to have two queues with the same service if the location is different?  (We don't need to support this yet if we aren't planning on having such a thing in our implementations, but should log an error.)

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
mogoodrich, line 20: Let's also add some quick comments of what this class does/doesn't do, see some of my comments below, ie note that it assumes that a service concept maps to one and only one queue.

mogoodrich, line 20: Sorry, one more thing :) can we add some unit tests around this?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+        Concept addToQueueConcept = Context.getConceptService().getConceptByMapping(ADD_TO_QUEUE_CONCEPT_CODE_PIH, "PIH");
+
+        if (addToQueueConcept == null) {
+            log.error("Concept PIH: " + ADD_TO_QUEUE_CONCEPT_CODE_PIH + " cannot be found");
+            return;
+        }
+        Patient patient = formEntrySession.getPatient();
+        Encounter encounter = formEntrySession.getEncounter();
+
+        for (Obs candidate : encounter.getObsAtTopLevel(false)) {
```
mogoodrich, line 37: Let's note in the comments I mentioned adding above that this action expects the queue question to be at the top level.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                    if (allowedStatuses.isEmpty()) {
+                        log.error("Queue " + queue.getName() + " has not valid allowed statuses");
+                        return;
+                    }
+                    List<Concept> allowedPriorities = queueServicesWrapper.getAllowedPriorities(queue);
+                    if (allowedPriorities.isEmpty()) {
+                        log.error("Queue " + queue.getName() + " has not valid allowed priorities");
+                        return;
+                    }
+                    queueEntry = new QueueEntry();
```
mogoodrich, line 61: Don't we have to test if the patient is already in the queue, or is this something that the queue entry service handles behind the scenes?  Specifically, what happens if you edit the form and re-save, does it add them to the queue again? Or are we assuming we aren't supporting the edit case yet?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                    queueEntry.setQueue(queue);
+                    queueEntry.setPatient(patient);
+                    // The first entry in the allowedPriorities should always be Normal
+                    queueEntry.setPriority(allowedPriorities.get(0));
+                    // The first entry in the allowedStatuses should always be Waiting
+                    queueEntry.setStatus(allowedStatuses.get(0));
+                    queueEntry.setStartedAt(new Date());
```
mogoodrich, line 68: I would think that this should be the encounter datetime instead of new Date() but I haven't 100% though through all the use cases with this yet, and it depends on whether we supporting retrospective entry to start out with.

mogoodrich (changes requested):
Thanks @cioan !  Yes, totally makes sense to me to have the value coded = the queue service so we can use this handler for any queue service type.  See my comments on the code re: some other thoughts/questions.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                        log.error("Queue " + queue.getName() + " has not valid allowed priorities");
+                        return;
+                    }
+                    queueEntry = new QueueEntry();
+                    queueEntry.setQueue(queue);
+                    queueEntry.setPatient(patient);
+                    // The first entry in the allowedPriorities should always be Normal
+                    queueEntry.setPriority(allowedPriorities.get(0));
+                    // The first entry in the allowedStatuses should always be Waiting
+                    queueEntry.setStatus(allowedStatuses.get(0));
```
mseaton, line 67: I don't think we should make this assumption.  I think we should just set priority to Normal and status to Waiting explicitly if that is what we want to see happen.  Then, if there are situations where we want something different, we can update the logic here explicitly to handle this.  This is in the pihcore module, and our metadata definition is in the pihemr config, we know what is valid and not valid, and we should fail hard if the expected situation is not found, as this reflects something wrong happened in the code when starting up.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                Queue queue = null ;
+                QueueEntry queueEntry = null;
+                if ( !queues.isEmpty() ) {
+                    queue = queues.get(0);
+                }
+                if (queue != null ) {
+                    List<Concept> allowedStatuses = queueServicesWrapper.getAllowedStatuses(queue);
+                    if (allowedStatuses.isEmpty()) {
+                        log.error("Queue " + queue.getName() + " has not valid allowed statuses");
+                        return;
```
mseaton, line 54: Per my comment below, I think we can remove this block of allowedStatuses entirely.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
```diff
@@ -0,0 +1,75 @@
...
+                    }
+                    List<Concept> allowedPriorities = queueServicesWrapper.getAllowedPriorities(queue);
+                    if (allowedPriorities.isEmpty()) {
+                        log.error("Queue " + queue.getName() + " has not valid allowed priorities");
+                        return;
```
mseaton, line 59: Per my comment below, I think we can remove this block of allowedPriorities entirely.

mseaton (changes requested):

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
cioan: np, the class has been renamed.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
cioan: yes, the QueueEntryService will throw an exception:
https://github.com/openmrs/openmrs-module-queue/blob/main/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java#L95

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
cioan: @mseaton , it is easy enough to just hard code the Status and Priority concepts, but we just did some configuration work on the queue module and the config-pihsl that allows us to configure each queue individually with different statuses and priorities. And you said that the best way to determine the allowed statuses and priorities for a given queue is to use the gerAllowed methods. We should ignore this possible custom configuration now, and just hardcode here the Normal and Waiting concepts? Then, maybe I should remove the serviceConcept too, and just hardcode to the MCOE Triage concept?

cioan:
Thanks @mseaton and @mogoodrich for your reviews. I think I addressed all your comments above. I am going to look into it how to create a test suite for this new class.

cioan:
> > I think I addressed all your comments above
> 
> Thanks @cioan . I resolved the things that seemed resolved. There are a few remaining comments that still need to be addressed I believe.

Thanks @mseaton ! Did you see the question about hard coding the Status, Priority and maybe the service concept?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueue.java
cioan: I just added unit tests. Thanks @mogoodrich !

cioan:
@mogoodrich and @mseaton , I have updated the PR with the changes we discussed yesterday:

- do not attempt to add patient to the queue if patient already on the queue
- add test cases for both scenarios

Thanks!

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
```diff
@@ -0,0 +1,104 @@
...
+                }
+                queueEntry.setStartedAt(encounter.getEncounterDatetime());
+
+                QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+                searchCriteria.setPatient(patient);
+                searchCriteria.setQueues(Collections.singletonList(queue));
+                List<QueueEntry> queueEntries = queueServicesWrapper.getQueueEntryService().getQueueEntries(searchCriteria);
+                boolean isPatientAlreadyOnTheQueue = false;
+                for (QueueEntry entry : queueEntries) {
+                    if (!entry.equals(queueEntry)) {
```
mseaton, line 87: This check is unnecessary - it will never be true as `queueEntry` is always a `new QueueEntry()`.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
```diff
@@ -0,0 +1,104 @@
...
+
+                List<Concept> allowedPriorities = queueServicesWrapper.getAllowedPriorities(queue);
+                if (!allowedPriorities.isEmpty()) {
+                    queueEntry.setPriority(allowedPriorities.get(0));
+                }
+                queueEntry.setStartedAt(encounter.getEncounterDatetime());
+
+                QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+                searchCriteria.setPatient(patient);
+                searchCriteria.setQueues(Collections.singletonList(queue));
```
mseaton, line 83: Tangentially, it would probably have been useful to have added an additional property to QueueEntrySearchCriteria (Date activeOnDate), to only return those queue entries from the DB, rather than have to iterate on the like we are doing below, but this is fine.

mseaton (approved):
Approving but see my comment about the unnecessary equals check

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
cioan: Thanks! It has been removed.

cioan:
Thanks @mseaton !

PR #478, opened 2024-02-14
UHM-7670 - Display error messages if a visit fails to save on edit.

mseaton:


mogoodrich (approved):
LGTM.  FWIW, I still think there's a bug in the code and it's supposed to stop you from changing to an invalid date, *but* good to have this error message regardless, and it's certainly arguable that allowing the use to make the change and displaying the reason why their change is invalid is clearer than just not allowing them to change the date to an invalid value without telling them why,

PR #480, opened 2024-02-21
UHM-7748, display only obs with values

cioan:
@mogoodrich , in this PR I have added another field to the template section that is an array of fields with values. So, before I made this change here is how the templateModel data looked like:

{
  "name": "Maternal Check-In",
  "fields": [
    {
      "name": "Reason for visit",
      "datatype": "coded",
      "value": "Family planning services"
    },
    {
      "name": "Patient referred to hospital",
      "datatype": "coded",
      "value": "Yes"
    },
    {
      "name": "Role of referring person",
      "datatype": "coded",
      "value": "Peripheral Health Unit"
    },
    {
      "name": "Other referral (text)",
      "datatype": "text",
      "value": ""
    },
    {
      "name": "Address of contact person",
      "datatype": "text",
      "value": ""
    },
    {
      "name": "Telephone number of contact",
      "datatype": "text",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": "Pain"
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": "Bleeding"
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reasons for referral (coded)",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Reason for referral (text)",
      "datatype": "text",
      "value": ""
    },
    {
      "name": "Type of referral ordered",
      "datatype": "coded",
      "value": ""
    },
    {
      "name": "Method of transportation",
      "datatype": "coded",
      "value": "Motorcycle"
    },
    {
      "name": "Transportation comment",
      "datatype": "text",
      "value": ""
    }
  ]
}

So I parsed this array of all the encounter obs, and added a separate node activeFields, where I included only the obs with values. And if an obs had multiple answers, then I combined those answers to one string comma delimited:
[
  {
    "name": "Reason for visit",
    "value": "Postnatal visit"
  },
  {
    "name": "Patient referred to hospital",
    "value": "Yes"
  },
  {
    "name": "Role of referring person",
    "value": "Traditional birth attendant (TBA)"
  },
  {
    "name": "Reasons for referral (coded)",
    "value": "Stage of labor, Antenatal visit, Other"
  },
  {
    "name": "Reason for referral (text)",
    "value": "doare"
  },
  {
    "name": "Type of referral ordered",
    "value": "Triage service"
  },
  {
    "name": "Method of transportation",
    "value": "Walking"
  }
]

Here is how the encounter was displaying before I made this change:

Screenshot 2024-02-21 at 10 00 02 AM

And here is how it looks now after adding the activeFields array of obs with values:
Screenshot 2024-02-21 at 10 23 01 AM

### DIFF
--- a/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
+++ b/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
@@ -11,17 +11,11 @@
 <div class="content encounter-summary-long">
 
     <span ng-repeat="section in templateModel.sections" class="aligned">
-        <span ng-repeat="field in (section.activeFields ? section.activeFields : section.fields)" class="aligned">
+        <span ng-repeat="field in section.activeFields" class="aligned">
             <p ng-show="field.value" class="aligned">
                 <label>{{ field.name }}</label>
                 <span class="obs-value">{{ field.value }}</span>
             </p>
-            <span ng-repeat="childField in field.fields" class="aligned">
-                <p ng-show="childField.value" class="aligned">
-                    <label>{{ (\$index == 0 || childField.name != field.fields[\$index-1].name) ? childField.name : '' }}</label>
-                    <span class="obs-value">{{ childField.value }}</span>
-                </p>
-            </span>
         </span>
     </span>
 
--- a/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
+++ b/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
@@ -11,7 +11,7 @@
 <div class="content encounter-summary-long">
 
     <span ng-repeat="section in templateModel.sections" class="aligned">
-        <span ng-repeat="field in section.activeFields" class="aligned">
+        <span ng-repeat="field in (section.activeFields ? section.activeFields : section.fields)" class="aligned">
             <p ng-show="field.value" class="aligned">
                 <label>{{ field.name }}</label>
                 <span class="obs-value">{{ field.value }}</span>
--- a/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
+++ b/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
@@ -11,11 +11,9 @@
 <div class="content encounter-summary-long">
 
     <span ng-repeat="section in templateModel.sections" class="aligned">
-        <span ng-repeat="field in section.fields" class="aligned">
+        <span ng-repeat="field in section.activeFields" class="aligned">
             <p ng-show="field.value" class="aligned">
-                <!-- this cryptic expression is to suppress label name if it is same as previous name -->
-                <!-- TODO: can we encapsulate this somewhere else -->
-                <label>{{ (\$index == 0 || field.name != section.fields[\$index-1].name) ? field.name : '' }}</label>
+                <label>{{ field.name }}</label>
                 <span class="obs-value">{{ field.value }}</span>
             </p>
             <span ng-repeat="childField in field.fields" class="aligned">
--- a/omod/src/main/webapp/resources/scripts/visit/visit.js
+++ b/omod/src/main/webapp/resources/scripts/visit/visit.js
@@ -123,6 +123,7 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                             $http.get("/" + OPENMRS_CONTEXT_PATH + url)
                                 .then(function (response) {
                                     $scope.templateModel = response.data;
+                                    getFieldsWithValue($scope.templateModel.sections);
                                     if ($scope.templateModel.html) {
                                         // this enabled the "viewEncounerWithHtmlFormLong" view to display raw html returned by the htmlformentryui module
                                         $scope.html = $sce.trustAsHtml($scope.templateModel.html);
@@ -138,7 +139,31 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                                     $scope.encounter = encounter;
                                 });
                         }
+                    }
 
+                    function getFieldsWithValue(sections) {
+                        if (sections) {
+                            sections.forEach((section) => {
+                                let activeFields = [];
+                                if (section.fields) {
+                                    for (let j = 0; j < section.fields.length; j++) {
+                                        const field = section.fields[j];
+                                        if (field.value) {
+                                            let foundField = activeFields.find(({ name }) => name === field.name);
+                                            if (foundField) {
+                                                foundField.value = foundField.value + ", " + field.value;
+                                            } else {
+                                                activeFields.push({
+                                                    name: field.name,
+                                                    value: field.value
+                                                });
+                                            }
+                                        }
+                                    }
+                                    section.activeFields = activeFields;
+                                }
+                            })
+                        }
                     }
 
                     $scope.goToLabResults = function () {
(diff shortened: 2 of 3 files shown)


omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
```diff
@@ -11,11 +11,9 @@
 <div class="content encounter-summary-long">
 
     <span ng-repeat="section in templateModel.sections" class="aligned">
-        <span ng-repeat="field in section.fields" class="aligned">
+        <span ng-repeat="field in section.activeFields" class="aligned">
             <p ng-show="field.value" class="aligned">
-                <!-- this cryptic expression is to suppress label name if it is same as previous name -->
-                <!-- TODO: can we encapsulate this somewhere else -->
-                <label>{{ (\$index == 0 || field.name != section.fields[\$index-1].name) ? field.name : '' }}</label>
```
mogoodrich, line 18: I think we want to remove all four instances of this "cryptic hack" in this file?

omod/src/main/webapp/resources/scripts/visit/visit.js
```diff
@@ -123,6 +123,7 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                             $http.get("/" + OPENMRS_CONTEXT_PATH + url)
                                 .then(function (response) {
                                     $scope.templateModel = response.data;
+                                    getFieldsWithValue($scope.templateModel.sections);
```
mogoodrich, line 126: I think we also need to handle forms with no sections, see: https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp#L30

omod/src/main/webapp/resources/scripts/visit/visit.js
```diff
@@ -138,7 +139,31 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                                     $scope.encounter = encounter;
                                 });
                         }
+                    }
 
+                    function getFieldsWithValue(sections) {
```
mogoodrich, line 144: Does this handle nested fields?

mogoodrich (commented):
Thanks @cioan . This is a good approach for this.  Generally looks good, though I think it need to be expanded to support nested obs groups (ie,what if a field has child fields as handled here: https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp#L21)

Also, it looks like in cases where a form does not have sections, it may have top-level fields, which we'd want to process in the same way (look likely use almost the same code), see: https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp#L30

You can see the various types of formats it supports via the test cases here:
https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/omod/src/test/java/org/openmrs/htmlformentry/web/controller/HtmlFormEncounterControllerTest.java

And here's one of the original tickets with comments about how this works:
https://openmrs.atlassian.net/browse/HTML-604

cioan:
> Thanks @cioan . This is a good approach for this. Generally looks good, though I think it need to be expanded to support nested obs groups (ie,what if a field has child fields as handled here: https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp#L21)
> 
> Also, it looks like in cases where a form does not have sections, it may have top-level fields, which we'd want to process in the same way (look likely use almost the same code), see: https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp#L30
> 
> You can see the various types of formats it supports via the test cases here: https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/omod/src/test/java/org/openmrs/htmlformentry/web/controller/HtmlFormEncounterControllerTest.java
> 
> And here's one of the original tickets with comments about how this works: https://openmrs.atlassian.net/browse/HTML-604

Thanks @mogoodrich ! I could go on and try to cover all the other use cases you pointed out. The nested fields and form with no sections, but after searching for a while through our forms and encounterTypeConfig.js I still cannot figure out which forms and encounters use the defaultHtmlFormEncounterLong template to display the obs on the Visit dashboard. Short of finding a real form that I could use to validate the changes I make, I am not sure how to go about testing and validating those changes in the defaultHtmlFormEncounterLong template? The test cases you pointed out in the HFE module may or may not correspond to all the forms we have in the PIH EMR. And those test cases test the structure of the form, not the template that we use to display those encounters and obs. 
I just added a small change to the code to fallback on the existing fields if the activeFields is missing. So, maybe this would be sufficient to guarantee that we are not breaking the display of any existing forms? 

So, please let me know if you could find any form that has no sections, or has child fields that use the defaultHtmlFormEncounterLong template and I would be happy to continue testing and making sure those encounter fields display correctly. Thanks!

mogoodrich:
It looks like the Haiti check-in form uses an obs group for the two payment obs, amount paid and receipt number, I just want to confirm you change doesn't break this use case:

cioan:
Thanks @mogoodrich ! I will test the Haiti check-in form.

cioan:
@mogoodrich , you were correct, the nested obs were no longer displayed. Here is how the fields, and the activeFields array look when using the Haiti check-in form:
Screenshot 2024-02-21 at 7 29 43 PM

The payment obs are not displayed:
Screenshot 2024-02-21 at 7 32 10 PM

I will address it.

cioan:
@mogoodrich , I got the nesting obs displaying correctly too. Here is the updated activeFields array after the last commit:

Screenshot 2024-02-22 at 7 50 31 AM

and here is the display on the visit dashboard:
Screenshot 2024-02-22 at 7 53 09 AM

omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
```diff
@@ -11,11 +11,9 @@
 <div class="content encounter-summary-long">
 
     <span ng-repeat="section in templateModel.sections" class="aligned">
-        <span ng-repeat="field in section.fields" class="aligned">
+        <span ng-repeat="field in (section.activeFields ? section.activeFields : section.fields)" class="aligned">
```
mogoodrich, line 14: For clarity, I'd be fine with just switching this to solely use the activeFields, with your new code I think if it ever got to "section.fields" is would be empty anyways?

mogoodrich (approved):
LGTM, I did make one comment, and in conjunction with that I would also remove lines 21-26, since you now also flatten the child obs in your new code so I don't think that child section really does much anymore.  Thanks!

omod/src/main/webapp/pages/visit/templates/encounters/defaultHtmlFormEncounterLong.gsp
cioan: Thanks @mogoodrich ! I have removed the lines you indicated.

mogoodrich:
Thanks @cioan !

PR #481, opened 2024-02-23
UHM-7716: prevent adding queue entries for past visits

cioan:
@mogoodrich , while testing this MCOE check-in form, I have noticed a strange error. If I go back to a past visit which had MCOE check-in encounter with Refer to MCOE Triage check-in obs, and try to edit the encounter then the postsubmit action is executed again and another queue entry is created on the past visit:

Screenshot 2024-02-23 at 9 43 03 AM

When I click to edit the MCOE Check-in encounter, the Refer to MCOE triage question is not displayed, because of this code:
https://github.com/PIH/openmrs-config-pihsl/blob/master/configuration/pih/htmlforms/checkinMaternal.xml#L259

            <ifMode mode="ENTER" include="true">
                <includeIf velocityTest="($visit.open)">
                    
                    <fieldset class="queue">
                        <legend>
                            <uimessage code="MCOE Triage"/>
                        </legend>
                        <h3>
                            <uimessage code="MCOE Triage"/>
                        </h3>

                        <p class="left">
                            <obs conceptId="CIEL:1272" style="checkbox" id="referToMCOETriage"
                                 answerConceptId="PIH:14976" answerCode="Refer to MCOE triage" />
                        </p>
                    </fieldset>
                </includeIf>
            </ifMode>

But the postSubmissionAction is executed, and because other problems in the queue module it would create duplicate queue entries. 

I even try something like this in the check-in form, but it did not work:

```
<ifMode mode="ENTER" include="true">
        <includeIf velocityTest="($visit.open)">
            <postSubmissionAction class="org.openmrs.module.pihcore.htmlformentry.action.AddPatientToQueueAction"/>
        </includeIf>
</ifMode>
```

The post submit action was included and executed on the edit mode, even if modified the form. I guess it retrieves the version of the form that was used to create the encounter initially? So, maybe for any new encounter the little code above would work?

Anyway, for now I thought I would modify the AddPatientToQueueAction postsubmitaction to not execute if the encounter is part of a visit that was closed. Does this make sense? Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
@@ -54,31 +54,13 @@ public class AddPatientToQueueAction implements CustomFormSubmissionAction {
                     criteria.setServices(Arrays.asList(serviceQueue));
                     QueueServicesWrapper queueServicesWrapper = Context.getRegisteredComponents(QueueServicesWrapper.class).get(0);
                     List<Queue> queues = queueServicesWrapper.getQueueService().getQueues(criteria);
-                    Queue queue = null;
-                    QueueEntry queueEntry = null;
                     if (queues.isEmpty()) {
                         throw new RuntimeException("AddPatientToQueueAction not executing as no queue with service concept " + serviceQueue.getName() + " could be found ");
                     }
                     if (queues.size() > 1) {
                         throw new RuntimeException("AddPatientToQueueAction not executing as there are more than one queue with service concept " + serviceQueue.getName());
                     }
-                    queue = queues.get(0);
-
-                    queueEntry = new QueueEntry();
-                    queueEntry.setQueue(queue);
-                    queueEntry.setPatient(patient);
-
-                    List<Concept> allowedStatuses = queueServicesWrapper.getAllowedStatuses(queue);
-                    if (!allowedStatuses.isEmpty()) {
-                        queueEntry.setStatus(allowedStatuses.get(0));
-                    }
-
-                    List<Concept> allowedPriorities = queueServicesWrapper.getAllowedPriorities(queue);
-                    if (!allowedPriorities.isEmpty()) {
-                        queueEntry.setPriority(allowedPriorities.get(0));
-                    }
-                    queueEntry.setStartedAt(encounter.getEncounterDatetime());
-                    queueEntry.setVisit(encounter.getVisit());
+                    Queue queue = queues.get(0);
 
                     QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
                     searchCriteria.setPatient(patient);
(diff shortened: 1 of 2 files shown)


mogoodrich:
Thanks @cioan !

Interesting, I would have guessed that your attempt at wrapping the post submission action in an ifMode would have worked.

As for the logic you added, wouldn't that still cause a problem if you edit a check in form while the patient is within an active visit?

Note that you can access the form entry context within the submission action, so the best/easiest thing to do would probably be to just skip the action if not in ENTER mode.  ie implement the equivalent of your ifMode stuff within the action itself.   To ge tthe mode:

`     formEntrySession.getContext().getMode().equals(FormEntryContext.Mode.ENTER)
`

Though another question I'd have, it looks like there is logic in the post submission handler that supposed to protect against adding a patient to a queue twice, is there a reason this is not handling this use case?

cioan:
Thanks @mogoodrich ! I have added the check for the form to be in the ENTER mode. I have also modified the code for searching the queue entries to find even old close queue entries. And in a separate PR, https://github.com/openmrs/openmrs-module-queue/pull/55 we are going to address this problem in the queue module. Thanks!

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
```diff
@@ -38,72 +36,68 @@ public class AddPatientToQueueAction implements CustomFormSubmissionAction {
...
-                }
-                if (!isPatientAlreadyOnTheQueue) {
-                    try {
-                        queueServicesWrapper.getQueueEntryService().saveQueueEntry(queueEntry);
-                    } catch (Exception e) {
-                        throw new RuntimeException("AddPatientToQueueAction failed to add patient to queue", e);
+                    QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+                    searchCriteria.setPatient(patient);
+                    searchCriteria.setQueues(Collections.singletonList(queue));
+                    searchCriteria.setIsEnded(null);
```
mseaton, line 86: Yes, this is currently needed and not great as you point out.  I have a PR in the queue module to address this so this should not be needed after / assuming that is merged.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
```diff
@@ -38,72 +36,68 @@ public class AddPatientToQueueAction implements CustomFormSubmissionAction {
...
-        }
-        Patient patient = formEntrySession.getPatient();
-        Encounter encounter = formEntrySession.getEncounter();
-        Visit visit = encounter.getVisit();
-        if ( visit != null && visit.getStopDatetime() != null) {
-            // this post submit action do not apply to retrospective entries, only for active point of care visits
-            log.warn("AddPatientToQueueAction not executing because this is not an active visit");
-            // we should not throw an error, because the user should be able to edit other obs on this check-in encounter
-            return;
-        }
```
mseaton: Why are you removing the above block @cioan ?  It's still possible to have a non-active visit in Enter mode.  Do we not want this check any more?

cioan: If the visit is not active then the Refer to triage question is not included in the form:

https://github.com/PIH/openmrs-config-pihsl/blob/master/configuration/pih/htmlforms/checkinMaternal.xml#L259

But, I could add back this code if you think that would be safer?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
```diff
@@ -37,65 +36,68 @@ public class AddPatientToQueueAction implements CustomFormSubmissionAction {
...
-                if (!isPatientAlreadyOnTheQueue) {
-                    try {
-                        queueServicesWrapper.getQueueEntryService().saveQueueEntry(queueEntry);
-                    } catch (Exception e) {
-                        throw new RuntimeException("AddPatientToQueueAction failed to add patient to queue", e);
+                    QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+                    searchCriteria.setPatient(patient);
+                    searchCriteria.setQueues(Collections.singletonList(queue));
+                    searchCriteria.setIsEnded(null);
+                    List<QueueEntry> queueEntries = queueServicesWrapper.getQueueEntryService().getQueueEntries(searchCriteria);
```
mogoodrich, line 87: This is a nit, but it seems odd to me that you instantiate the new QueryEntry and fill out all the parameter *before* actually doing the check to see if you need to save it.  Not really a big deal, but maybe add a comment here "// only save queue entry if patient not on queue" just for readability... I was a little confused because after line 67 would have assumed that the decision to save the queue had already been made, and a comment would highlight that it hadn't.  Hope that makes some sense!

mogoodrich (approved):
looks good to me, thanks @cioan , just a comment nit... :)

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/AddPatientToQueueAction.java
cioan: Thanks @mogoodrich ! I have refactored the code, and moved the code you pointed out in the same block as saving the new queue entry.

PR #482, opened 2024-02-27
SL-511 Change concept mapping for date of death for reporting

lnball:
Use the generic mapping for Date of Death.  This will be useful when we move to datetime.  I believe the same code will work for datetime (as it does now for date).


mogoodrich (approved):
Looks right to me, thanks @lnball !

PR #483, opened 2024-02-28
UHM-7768: Close stale visits task needs to account for queue entries …

mseaton:
…in addition to encounters


mseaton:
> Tangentially, it occurs to me that we need to close out any open queue entries when we close a visit, but I think this happens elsewhere?

Yes @mogoodrich , see UHM-7412

PR #484, opened 2024-02-29
UHM-7774: VelocityTests failing

mogoodrich:


api/src/test/java/org/openmrs/module/pihcore/status/StatusDataFunctionsTest.java
```diff
@@ -19,20 +19,30 @@ public void testVelocityExpressions() {
...
+
+        cal = Calendar.getInstance();
         cal.add(Calendar.DATE, 52);
         dataValues.put("in52Days", cal.getTime());
-        cal.add(Calendar.DATE, -52);
+
+        cal = Calendar.getInstance();
         cal.add(Calendar.DATE, -1);
         dataValues.put("yesterday", cal.getTime());
         String yesterday = df.format(cal.getTime());
```
mogoodrich, line 34: When run on leap day (Feb 29th) "yesterday" was getting set to Feb 27 2024 (not Feb 28th) because we were reusing a calendar instance, and so when jumping three years again from Feb 29, 2024 the calendar ends up on Feb 28.2027, and so when you subtract 3 years you end up on Feb 28, 2024, not Feb 29. 2024.

I changed these tests to instantiate a new calendar for each new element.

api/src/test/java/org/openmrs/module/pihcore/status/StatusDataFunctionsTest.java
```diff
@@ -19,20 +19,30 @@ public void testVelocityExpressions() {
...
-        cal.add(Calendar.DATE, -7);
```
mogoodrich, line 31: Note that this was subtracting 7 days from yesterday, but I changed it to subtract a week from today.

api/src/test/java/org/openmrs/module/pihcore/status/StatusDataFunctionsTest.java
```diff
@@ -41,7 +51,7 @@ public void testVelocityExpressions() {
         testExpression(dataValues, today, "$fn.formatDate($fn.now(), 'yyyy-MM-dd')");
         testExpression(dataValues, "0", "$fn.monthsSince($lastWeek)");
         testExpression(dataValues, "1", "$fn.monthsSince($lastMonth)");
-        testExpression(dataValues, "13", "$fn.monthsSince($lastYear)");
+        testExpression(dataValues, "12", "$fn.monthsSince($lastYear)");
```
mogoodrich, line 54: Since we are starting with an new instance (and not one that is 8 days behind) this becomes 12 months, not 13 months, which seems more correct?  It's possible that there was a reason for doing it this way though, and we are going to start running into issues on other dates?

cioan (approved):
Thanks @mogoodrich !

mseaton (approved):
LGTM

mogoodrich:
Thanks all! Merging this in...

PR #485, opened 2024-03-06
UHM-7787 Change lookup for Date of Death to use PIH mapping rather th…

mseaton:
…an emrapi mapping in pihcore


lnball:
Those mappings point to the same concept....  How does this change anything?

PR #486, opened 2024-03-07
UHM-7779: Make OCL the default/only option for concept loading

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/config/PihConfigService.java
```diff
@@ -94,20 +92,6 @@ public InitializerConfig getInitializerConfig() {
...
-        if (mdsFiles != null) {
-            for (File f : mdsFiles) {
-                ConfigFile mdsFile = new ConfigFile();
-                mdsFile.setPath(f.getName());
-                mdsFile.setEnabled(true);
-                pihConceptsDomain.getConfigFiles().add(mdsFile);
-            }
-        }
-        return pihConceptsDomain;
-    }
```
mogoodrich, line 109: @mseaton  I wasn't 100% sure what this was used for, can we confirm it's not longer needed?

mseaton (approved):
Great work, I love this kind of cleanup :)

lnball (approved):
Very satisfying cleanup.  Does this mean we don't use mds module anymore?

mogoodrich:
@lnball , yes, it should be going away from the distro as part of this change.

Thanks, merging!

PR #487, opened 2024-03-07
SL-553: Add MCOE triage form and visit action

cioan:
Added MCOE Triage visit action.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
@@ -148,7 +148,7 @@ public class CustomAppLoaderConstants {
     public static final class Extensions {
 
         public static final String CHECK_IN_VISIT_ACTION = "pih.checkin.visitAction";
-        public static final String TRIAGE_VISIT_ACTION = "pih.triage.visitAction";
+        public static final String MCH_TRIAGE_VISIT_ACTION = "pih.mchtriage.visitAction";
         public static final String CHECK_IN_MATERNAL_VISIT_ACTION = "pih.checkinMaternal.visitAction";
         public static final String CHECK_IN_REGISTRATION_ACTION = "pih.checkin.registrationAction";
         public static final String CHECK_IN_MATERNAL_REGISTRATION_ACTION = "pih.checkinMaternal.registrationAction";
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
```diff
@@ -148,6 +148,7 @@ public static final class Apps {
     public static final class Extensions {
 
         public static final String CHECK_IN_VISIT_ACTION = "pih.checkin.visitAction";
+        public static final String TRIAGE_VISIT_ACTION = "pih.triage.visitAction";
```
lnball, line 151: There are other triage_visits and this one seems specific for mch triage.  If you don't think it's needs, I'm fine with that @cioan

lnball (commented):
See comment

mogoodrich (approved):
One small comment, but generally LGTM, thanks @cioan !

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
mogoodrich: Yes, agreed, that  including "MCOE" in the name males sense.

mogoodrich: Or perhaps "maternal" or "MCH" triage if we want something more generic?

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -231,6 +231,10 @@ private void loadAppsAndExtensions() throws UnsupportedEncodingException {
             enableCheckIn(config);
         }
 
+        if (config.isComponentEnabled(Components.MCOE_TRIAGE)) {
```
mogoodrich, line 234: I'm wondering if we want to name this something like "maternal triage" or "mch triage"?

mogoodrich (approved):
Actually, see my updated thoughts... I'm wondering if we want to try to use "MCH" and/or "maternal" as much as possible so we don't tie the naming convention to a specific facility in the specific country?  I guess it depends on how much we think this form is MCOE-specific (would we ever use it at Wellbody?)

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
cioan: Thanks! I have renamed to MCH triage.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
cioan: Thanks!

PR #488, opened 2024-03-08
UHM-7779: Make OCL the default/only option for concept loading

mogoodrich:
(disable loading metadata in a separate thread)

@mseaton  I doubt you will have issue with this... :)


mseaton (approved):
This is fine as-is for what it is doing, but we should also now review the "configureNonConceptDependencies" and "configureConceptDependencies" methods, eliminate the need for both of them, and instead just install our domains in the right order, either by just deferring to Iniz, or by doing it in one place.

mogoodrich:
Fair enough, let me work on this and test.

mogoodrich (commented):
@mseaton  can you re-review, I have combined the non-concept and concept methods into a single dependency method.

Note I reordered things a bit... OCL and post-concepts seems like they could/should now go further up in the startup process.  it turned out that reports had to go a bit later on because they relied on some mappings being set up.

I also noticed that I don't believe we need the custom drug loader functionality anymore?  I'm testing one more time with that removed, will not merge until I confirm it still works.

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
```diff
@@ -174,6 +171,14 @@ public void configureNonConceptDependencies() throws Exception {
         setStatus("Loading initializer pre-concept domains");
         InitializerSetup.loadPreConceptDomains(config);
 
+        // install concepts from OCL
+        setStatus("Installing Concepts from OCL zip");
+        InitializerSetup.installDomain(OCL, config);
+
+        // Load remaining Initializer domains that could depend on Concepts
+        setStatus("Loading initializer post-concept domains");
+        InitializerSetup.loadPostConceptDomains(config);
```
mseaton, line 180: At this point, we don't need these 3 steps at all @mogoodrich .  We should just call a single `InitializerSetup.install(config)` method that just installs from Iniz in the default order that Iniz specifies (we don't need any custom domain arrays to load in a specific order), and just applies our custom logic about site-specific configs and re-loading some domains if others have changed (eg. location / location tag).

Basically, convert `loadPreConceptDomains` into an `install` method, and remove the steps that specifically exclude `ocl` or anything in the `getDomainsToLoadAfterConcepts` list, which can be totally removed, but which keeps the logic that excludes files meant for a specific site.

mseaton (changes requested):

mseaton:
> Note I reordered things a bit... OCL and post-concepts seems like they could/should now go further up in the startup process. it turned out that reports had to go a bit later on because they relied on some mappings being set up

See my comments.  We shouldn't need to specify ordering at all anymore.  Let Iniz worry about that now.

mseaton:
> I also noticed that I don't believe we need the custom drug loader functionality anymore? I'm testing one more time with that removed, will not merge until I confirm it still works.

Correct, we just use Iniz for drugs.

api/src/main/java/org/openmrs/module/pihcore/setup/ConfigurationSetup.java
mogoodrich: Right, duh, thanks @mseaton  for the step-by-step details of what to change.  I just reworked and confirmed that a HUMCI SDK installed without error.  Will test the others after your review.

mseaton (approved):
This looks great @mogoodrich - thanks for doing this clean up!

mogoodrich:
Thanks @mseaton , merging this in!

PR #489, opened 2024-03-13
SL-553: add ability to set queue entry priority

cioan:
I just added to the existing AddPatientToQueueAction the ability to set the queue entry priority via another hidden coded question that could be set automatically based on the triage symptoms.


cioan:
Thanks @mseaton !

PR #490, opened 2024-03-22
UHM-7757: View Patient Appointments Via O2 Patient Chart

mogoodrich:


omod/src/main/webapp/fragments/dashboardwidgets/appointments.gsp
```diff
@@ -0,0 +1,29 @@
+<div class="info-section">
+    <div class="info-header">
+        <i class="${app.icon}"></i>
+        <h3>${ ui.message(app.label).toUpperCase() }</h3>
+        <a href="${ "/" + contextPath + "/spa/home/appointments/patient/" +   patientUuid }" class="right">
```
mogoodrich, line 5: This won't work until the new patient appointments appointments app functionality is merged in

omod/src/main/webapp/fragments/dashboardwidgets/appointments.gsp
```diff
@@ -0,0 +1,29 @@
...
+    <div class="info-body">
+        <ul>
+            <% appointments.each { appointment -> %>
+            <li>
+                <div style="display: block; overflow: hidden; padding-right: 5px; padding-bottom: 2px;">
+                    <div style="float: left">
+                        ${ui.formatDatetimePretty(appointment.startDateTime)}
+                        <br/>
+                        <strong>
+                            ${ appointment.service?.name }
```
mogoodrich, line 18: Unfortunately, AppointmentServiceDefinition is OpenmrsData, not OpenmrsMetadata, so our standard ${ui.format(appointment.service} pattrn doesn't work, though assumedly it would be trivial to add similar support for OpenmrsData to the UI Formatter if we think that is "correct".

omod/src/main/webapp/fragments/dashboardwidgets/appointments.gsp
```diff
@@ -0,0 +1,29 @@
...
+                        </strong>
+                    </div>
+                    <div  style="float: right; vertical-align: middle;">
+                        ${  ui.message("pihcore.appointments.status." + appointment.status) }
```
mogoodrich, line 22: added these codes to the PIH EMR

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/AppointmentsFragmentController.java
```diff
@@ -0,0 +1,76 @@
...
+                patient = patientService.getPatientByUuid((String)patientConfig);
+            }
+        }
+        if (patient == null) {
+            throw new IllegalArgumentException("No patient found. Please pass patient into the configuration");
+        }
+
+        AppointmentSearchRequest appointmentSearchRequest = new AppointmentSearchRequest();
+        appointmentSearchRequest.setPatientUuid(patient.getUuid());
+        appointmentSearchRequest.setStartDate(new DateTime().withTimeAtStartOfDay().toDate());
```
mseaton, line 56: So we are explicitly not showing appointments scheduled for any day prior to the current day, but which still have a status indicating they have not been honored?

omod/src/main/webapp/fragments/dashboardwidgets/appointments.gsp
mseaton: I don't think that would work, because OpenmrsData does not have a `name` property, so you couldn't do anything with name.  You'd have to use reflection to support anything with a "name" property, that might work.  But I think this is fine as-is.  If we need this to be localized, we can figure that out and code that in outside of the ui.format tag.

mseaton (approved):
LGTM, pending the answer to my question about exactly what appointment we want to appear here.  Should they be all appointments that have a status of "SCHEDULED", regardless of date?  All appointments from today on regardless of status?  etc.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/AppointmentsFragmentController.java
mogoodrich: Good points, was just going based on the ticket description, will flag you there...

mogoodrich:
Merging this in to get the base functionality in place, but I have the overall design question regarding which appointments to display added for discussion on the ticket.

PR #491, opened 2024-04-01
SL-562, add nextSection parameter

cioan:


omod/src/main/webapp/resources/scripts/visit/visit.js
```diff
@@ -630,7 +630,7 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
 
             }
 
-            function goToNextSection(currentSection, patientUuid, encounterUuid, visitUuid) {
+            function goToNextSection(currentSection, nextSection, patientUuid, encounterUuid, visitUuid) {
```
mogoodrich, line 633: Can we add a comment here saying what this function does, because I'm sure I'll get confused by exactly what "next" means when I look at this a few years from now. :)

Just to make sure I have it correct, its:
* If "nextSection" has a value, jump to that specific section
* otherwise, if "nextSection" is null, jump to the next section in chronological sequence based on the encounter type config

mogoodrich (approved):
Thanks @cioan , looks great, just one note about adding a comment because I know I'm going to forgot the different things we are trying to do here

PR #496, opened 2024-05-13
SL-613, list patient's children

cioan:
Just listing patient's children, as they are defined by the Mother-to-Child relationship:

Screenshot 2024-05-13 at 2 48 50 PM


cioan:
Thanks @mogoodrich !

PR #500, opened 2024-05-30
SL-614, register child using mother's info

cioan:
@mogoodrich , I am just using the existing format SECTION.QUESTION.FIELD as it is defined in the [SectionsDefault.java](https://github.com/PIH/openmrs-module-pihcore/blob/master/api/src/main/java/org/openmrs/module/pihcore/apploader/apps/patientregistration/SectionsDefault.java#L60) to pass the mother's info to the registration form. 
When the registration form is loaded the fields that have initial values are complete and the corresponding question is checked as complete:
Screenshot 2024-05-30 at 7 38 02 AM


mogoodrich (approved):
Sweet, thanks @cioan !

PR #503, opened 2024-06-06
UHM-8012 - Upgrade to coreapps 2.0.0-SNAPSHOT and change code to acco…

mseaton:
…unt for backwards incompatible changes.


mogoodrich (approved):
Not 100% following, but I assume moving the activator and other files from api to omod was intentional as they utilized some web-only functionality?

PR #506, opened 2024-06-17
UHM-8038 - migrate fingerprint data into haitihivtest

mseaton:


mogoodrich (approved):
@mseaton  is making a change to search on both HIV EMR ID and ZL EMR ID in one search to see what duplicated we get, but otherwise looks good!

PR #507, opened 2024-07-11
SL-618: register baby with the initial data captured via the labour a…

cioan:
…nd delivery form

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
@@ -1,0 +1,55 @@
+package org.openmrs.module.pihcore.action;
+
+import java.util.Map;
+import org.apache.commons.lang.StringUtils;
+import org.openmrs.Concept;
+import org.openmrs.Obs;
+import org.openmrs.Patient;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.ObsService;
+import org.openmrs.module.registrationapp.action.AfterPatientCreatedAction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+
+@Component("registerBaby")
+public class RegisterBaby implements AfterPatientCreatedAction {
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    private static String YES_CONCEPT = "1065";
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
```diff
@@ -0,0 +1,55 @@
...
+    public void afterPatientCreated(Patient patient, Map<String, String[]> map) {
+        Concept noConcept = conceptService.getConceptByMapping(NO_CONCEPT, "CIEL");
+        Concept yesConcept = conceptService.getConceptByMapping(YES_CONCEPT, "CIEL");
+
+        if (map != null && !map.isEmpty()) {
+            String[] returnUrl = map.get("returnUrl");
+            if (returnUrl != null && returnUrl.length > 0) {
+                String url = returnUrl[0];
+                if (StringUtils.isNotBlank(url)) {
+                    String obsUuid = url.substring(url.indexOf("registerBabyObs=") + "registerBabyObs=".length());
```
mogoodrich, line 40: A bit funky to rely on the return url, but cool that it works!  Can we add a short overall comment to this class just explaining what it is doing since years down the road we will likely forget?

api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
```diff
@@ -0,0 +1,55 @@
...
+                    if (StringUtils.isNotBlank(obsUuid)) {
```
mogoodrich, line 41: Iguess there's also a slight risk that if this gets changed so that "registerBabyObs" isn't the last element in the returnUrl this will stop working? Should we change substriing to just grab a uuid-length instead of the rest of the string?

mogoodrich (approved):
A couple comments but LGTM thanks @cioan

api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
cioan: Makes sense @mogoodrich ! I have updated the code.

api/src/main/java/org/openmrs/module/pihcore/action/RegisterBaby.java
cioan: Sure, I have added a few comments to the top of the class.

cioan:
Thanks @mogoodrich ! I have also added the ability to navigate directly to the Print ID card as it is requested in the SL-618.

PR #512, opened 2024-07-29
SL:698: add post submission action to automatically register the deli…

cioan:
…very of a baby


api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -0,0 +1,192 @@
...
+        if (relationship !=null) {
+            relationships.add(new Relationship(mother, null, motherToChildRelationshipType));
+        }
+        Patient baby = new Patient();
+        PersonName babyName = new PersonName();
+        babyName.setFamilyName(mother.getFamilyName());
+        babyName.setGivenName("Baby");
+        baby.addName(babyName);
+        String babyGender = gender;
+        if (StringUtils.equalsIgnoreCase(gender, "Female")) {
```
mogoodrich, line 106: Can we change to match directly on concept instead of valueAsString?  (I think valueAsString is locale/language specific, among other things)?

mogoodrich (changes requested):
One change requested, but generally LGTM @cioan !

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: that is what I initially did, but the concept used here https://kgh-test.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=3253 returns "Male". If we store "Male" as gender in the db, that would be different than all other records,  which are stored as "M".
Would that be OK to store "Male" as a gender in the db?

mseaton: What @mogoodrich is talking about is line 68 @cioan , where you get "Male" as a string obs value, rather than getting `3ce84b1c-26fe-102b-80cb-0017a47871b2` as the obs value coded concept id.  Please do that instead.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    Patient registerBaby(Patient mother, String gender, Date birthDatetime, Location location) {
+        PersonService personService = Context.getPersonService();
+        RelationshipType motherToChildRelationshipType = personService.getRelationshipTypeByUuid(PihEmrConfigConstants.RELATIONSHIPTYPE_MOTHERTOCHILD_UUID);
+        RelationshipType relationship = personService.getRelationshipTypeByUuid(motherToChildRelationshipType.getUuid());
+        List<Relationship> relationships = new ArrayList<Relationship>();
+        if (relationship !=null) {
+            relationships.add(new Relationship(mother, null, motherToChildRelationshipType));
```
mseaton, line 98: I don't understand this.  We are creating a relationship before creating the baby?  And leaving the personB null?  That doesn't seem right.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -0,0 +1,192 @@
...
+            //automatically redirect to the children page
+            formEntrySession.setAfterSaveUrlTemplate("pihcore/children/children.page?patientId={{patient.id}}");
+        }
+    }
+
+    Patient registerBaby(Patient mother, String gender, Date birthDatetime, Location location) {
+        PersonService personService = Context.getPersonService();
+        RelationshipType motherToChildRelationshipType = personService.getRelationshipTypeByUuid(PihEmrConfigConstants.RELATIONSHIPTYPE_MOTHERTOCHILD_UUID);
+        RelationshipType relationship = personService.getRelationshipTypeByUuid(motherToChildRelationshipType.getUuid());
+        List<Relationship> relationships = new ArrayList<Relationship>();
```
mseaton, line 96: Are there circumstances we need to handle here where the obs says that the baby hasn't been registered yet, but in fact they have?  Specifically for retrospective entry?  Do we want to sanity check that the mother doesn't have an existing relationship to a baby with the same birthdate before creating a new patient here?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
mogoodrich: Yes, correct, @mseaton

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: @mseaton , i just follow the same logic/workflow from registrationapp:
https://github.com/openmrs/openmrs-module-registrationapp/blob/master/omod/src/main/java/org/openmrs/module/registrationapp/fragment/controller/RegisterPatientFragmentController.java#L168

That could be incorrect too, but I don't think any relationships are actually saved until the patient is saved?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: np, I updated the code

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -0,0 +1,192 @@
...
+        String babyGender = gender;
+        if (StringUtils.equalsIgnoreCase(gender, "Female")) {
+            babyGender = "F";
+        } else if (StringUtils.equalsIgnoreCase(gender,"Male")){
+            babyGender = "M";
+        }
+        baby.setGender(babyGender);
+        baby.setBirthdate(birthDatetime);
+        PersonAddress motherAddress = mother.getPersonAddress();
+        if ( motherAddress != null ) {
```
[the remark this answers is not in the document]
cioan: ok, i've changed it to clone the mother's address

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: I thought we discussed those problems last week, and we decided to redirect the user to the children page where they will be able to remove any duplicates in the system, or search for another patient record?

cioan:
Merging this in. Happy to address any remaining post-commit comments.

PR #516, opened 2024-08-14
UHM-7443 and UHM-7992 - pregnancy program dashboard

mseaton:


api/src/main/java/org/openmrs/module/pihcore/apploader/ConfigAppLoaderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+                            files.add(f);
+                        }
+                    }
+                }
+            }
+        }
+        return files;
+    }
+
+    @Override
```
mseaton, line 61: This loads any file under `configuration/appframework/*/*AppTemplates.json`

api/src/main/java/org/openmrs/module/pihcore/apploader/ConfigAppLoaderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+            catch (Exception e) {
+                logger.error("Error reading AppTemplates configuration file: {}", f.getName(), e);
+            }
+        }
+        return templates;
+    }
+
+    @Override
+    public List<AppDescriptor> getAppDescriptors() {
```
mseaton, line 77: This loads any file under `configuration/appframework/*/*app.json`.  This also does some magic to avoid some boilerplate in the app and extensions definitions, namely it sets app ids and extension ids if they are not already set

api/src/main/java/org/openmrs/module/pihcore/apploader/ConfigAppLoaderFactory.java
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+            catch (Exception e) {
+                logger.error("Error reading AppDescriptors configuration file: {}", f.getName(), e);
+            }
+        }
+        return descriptors;
+    }
+
+    @Override
```
mseaton, line 116: I don't expect we will actually use this, but if we need to have free-standing extensions, this will load an file under `configuration/appframework/*/*extension.json`

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -3731,6 +3731,17 @@ private void enablePrograms(Config config) {
     }
 
     private void configureBasicProgramDashboard(String programUuid) {
```
mseaton, line 3733: All this is doing is saying - do not add any apps or extensions for the Sierra Leone Pregnancy Patient Dashboard, but otherwise continue to do so.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/EncountersFragmentController.java
```diff
@@ -86,19 +92,42 @@ else if (patientConfig instanceof String) {
 		}
 		model.put("encounterTypeToUrlMap", encounterTypeToUrlMap);
```
mseaton, line 94: This looks like there are more changes than there really are as I moved some things around, but this just adds support for an additional (optional) configuration property that will limit the returned encounters to being on or after the latest enrollment date of the given program.  If the patient does not have an active enrollment, then no encounters are returned.  If this configuration is not provided, then this should have no change from before.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -3731,6 +3731,17 @@ private void enablePrograms(Config config) {
...
+        boolean isSierraLeone = config.getCountry().equals(ConfigDescriptor.Country.SIERRA_LEONE);
+        boolean isPregnancyProgram = PihEmrConfigConstants.PROGRAM_PREGNANCY_UUID.equals(programUuid);
+        // The Sierra Leone Pregnancy Program Patient Dashboard is setup via config-pihsl, so do not set that up here
+        if (!isSierraLeone || !isPregnancyProgram) {
+            configureBasicProgramPatientDashboardWidgets(programUuid);
+        }
+        configureBasicProgramSummaryApps(programUuid);
+        configureBasicProgramExtensions(programUuid);
```
mogoodrich, line 3741: Shouldn't these be within the above if block?  I assume we aren't configuring part of the dashboard via the CALF and part within config-pihsl? Or am I missing something?

mseaton: These are different things than the clinician-facing dashboard, so I left them separate for now.

mogoodrich: Got it, thanks @mseaton

PR #518, opened 2024-08-22
SL-701: create visit and admission encounter for newly registered babies

cioan:
This seems to be working fine:

Screenshot 2024-08-22 at 4 42 19 PM

Screenshot 2024-08-22 at 4 42 00 PM

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
@@ -20,6 +20,7 @@ import org.openmrs.api.PersonService;
 import org.openmrs.api.VisitService;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.emrapi.adt.AdtService;
+import org.openmrs.module.emrapi.visit.VisitDomainWrapper;
 import org.openmrs.module.htmlformentry.CustomFormSubmissionAction;
 import org.openmrs.module.htmlformentry.FormEntryContext;
 import org.openmrs.module.htmlformentry.FormEntrySession;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -150,6 +161,54 @@ Patient registerBaby(Patient mother, Concept gender, Date birthDatetime, Locatio
...
+
+        EncounterService encounterService = Context.getEncounterService();
+        Encounter lastAdmissionEnc = null;
+        List<Encounter> motherEncounters =  motherEncounter.getVisit().getNonVoidedEncounters();
+        // look through mother's encounters for the most recent Admission encounter
+        for (Encounter enc : motherEncounters) {
+            if (enc.getEncounterType().getUuid().equalsIgnoreCase(PihEmrConfigConstants.ENCOUNTERTYPE_ADMISSION_UUID)) {
+                if ( lastAdmissionEnc == null ||
+                        ( (lastAdmissionEnc !=null) && lastAdmissionEnc.getEncounterDatetime().before(enc.getEncounterDatetime()) ) ) {
+                    lastAdmissionEnc = enc;
```
mogoodrich, line 187: We need to take into account transfers within the hospital here as well, so this needs to check transfer encounters as well.

That being said, there's also a utility method from EMR API you should be able to use as well, using a onDate of now, if you wrap the visit in the visit domain wrapper:

https://github.com/openmrs/openmrs-module-emrapi/blob/master/api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java#L515

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -150,6 +161,54 @@ Patient registerBaby(Patient mother, Concept gender, Date birthDatetime, Locatio
...
+            admissionEncounter = new Encounter();
+            admissionEncounter.setPatient(patient);
+            admissionEncounter.setEncounterType(encounterService.getEncounterTypeByUuid(PihEmrConfigConstants.ENCOUNTERTYPE_ADMISSION_UUID));
+            admissionEncounter.setVisit(visit);
+            admissionEncounter.setEncounterDatetime(birthDatetime);
+            admissionEncounter.setLocation(lastAdmissionEnc.getLocation());
+            Set<EncounterProvider> providers =  motherEncounter.getEncounterProviders();
+            if (providers != null && !providers.isEmpty()) {
+                for (EncounterProvider provider : providers) {
+                    admissionEncounter.addProvider(provider.getEncounterRole(), provider.getProvider());
```
mogoodrich, line 203: It's a good point about the provider, I'm not sure if it should be to the provider on the mother encounter, let me ping Fiona.

mogoodrich (changes requested):
Thanks @cioan ... one requested change, and one question for @fiona855 that I'll ask on the ticket.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: Thanks @mogoodrich ! I have updated to code to use the Visit Domain Wrapper to retrieve the mother's inpatient location. Thanks!

mogoodrich (approved):
LGTM, thanks @cioan !

cioan:
Thanks @mogoodrich !

PR #523, opened 2024-09-13
SL-667: Auto enrollment in the Pregnancy Program from a form

mogoodrich:
SL-669: Auto-update Program State from forms

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/PregnancyProgramEnrollmentAction.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/PregnancyProgramEnrollmentAction.java
@@ -1,0 +1,70 @@
+package org.openmrs.module.pihcore.htmlformentry.action;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.Encounter;
+import org.openmrs.Patient;
+import org.openmrs.PatientProgram;
+import org.openmrs.Program;
+import org.openmrs.ProgramWorkflowState;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.htmlformentry.CustomFormSubmissionAction;
+import org.openmrs.module.htmlformentry.FormEntryContext;
+import org.openmrs.module.htmlformentry.FormEntrySession;
+import org.openmrs.module.pihcore.SierraLeoneConfigConstants;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.openmrs.module.pihcore.htmlformentry.action.util.PregnancyProgramActionUtils.endEnrollment;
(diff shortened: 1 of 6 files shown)


mogoodrich:
A couple key points:

- This should work for the “real-time” use case and when entering forms retrospective but in order by encounter date.  Multiple programs enrollments will likely be created if the forms are entered retrospectively out of order.  We discussed this a bit during a standup this week but couldn’t come to a conclusion, so I implemented like this for now and we discuss other possibilities such as trying to move the enrollment date intelligently, or doing nothing when entering retrospectively 
-
- I implemented this so it does not do anything when editing a form.  The only reason this would/should do anything in “edit” mode would be if the encounter date was changed, and it would likely do wrong things similar to what is mentioned in the first bullet point (create duplicate enrollments).  We can discuss if we want more complex logic that tries to edit/change enrollment and state transition states when a form is edited and an encounter date is changed, but I’m hoping we can avoid this complexity and rely on “data cleanup” in these cases.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/PregnancyProgramEnrollmentAction.java
```diff
@@ -0,0 +1,70 @@
...
+    public void applyAction(FormEntrySession formEntrySession) {
+        // only do anything if we are entering (not editing) a form
+        if (formEntrySession.getContext().getMode().equals(FormEntryContext.Mode.ENTER)) {
+            Program pregnancyProgram = Context.getProgramWorkflowService().getProgramByUuid(SierraLeoneConfigConstants.PROGRAM_PREGNANCY_UUID);
+            ProgramWorkflowState antenatalState = Context.getProgramWorkflowService().getStateByUuid(SierraLeoneConfigConstants.PROGRAMWORKFLOW_PREGNANCYPROGRAMTYPEOFTREATMENT_STATE_ANTENATAL_UUID);
+
+            Patient patient = formEntrySession.getPatient();
+            Encounter encounter = formEntrySession.getEncounter();
+            // get all patient programs completed on or after encounter date (or not yet completed) and sort by date enrolled
+            List<PatientProgram> patientPregnancyPrograms = Context.getProgramWorkflowService().getPatientPrograms(patient, pregnancyProgram, null, null, encounter.getEncounterDatetime(), null, false);
```
cioan, line 47: @mogoodrich , how about the programs whose enrollment date is prior to the encounter date, and the dateCompleted is null? Would this query return those programs as well?

cioan (approved):
Looks good, @mogoodrich ! I just had a question, which became obsolete after I ready your comments on the other postsubmit action. Thanks!

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/PregnancyProgramEnrollmentAction.java
mogoodrich: From the javadoc, I believe the one date parameter I set should "if supplied will limit PatientPrograms to those completed on or after this Date OR not yet completed"

PR #524, opened 2024-09-19
SL-671: redirect to a new program summary page

cioan:
@mogoodrich , here is the only change I made to the CALF to point to the new program summary page.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
@@ -3804,7 +3804,7 @@ public class CustomAppLoaderFactory implements AppFrameworkFactory {
         apps.add(addToProgramSummaryListPage(app("pih.app." + programUuid + ".programSummary.dashboard",
                         "pih.app." + programUuid + ".programSummary.dashboard",
                         "fas fa-fw fa-list-alt",
-                        "/coreapps/summarydashboard/summaryDashboard.page?app=" + "pih.app." + programUuid + ".programSummary.dashboard",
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -3804,7 +3804,7 @@ private void configureBasicProgramSummaryApps(String programUuid) {
         apps.add(addToProgramSummaryListPage(app("pih.app." + programUuid + ".programSummary.dashboard",
                         "pih.app." + programUuid + ".programSummary.dashboard",
                         "fas fa-fw fa-list-alt",
-                        "/coreapps/summarydashboard/summaryDashboard.page?app=" + "pih.app." + programUuid + ".programSummary.dashboard",
```
mogoodrich, line 3807: got it... this works, though for clarity, let's just do a global replace in PIH Core of:

"programSummary.dashboard" -> "programEnrollments.dashboard"

and change the "addToProgramSummaryDashboardFirstColumn" to "addToProgramEnrollmentsDashboardFirstColumn"

mogoodrich (approved):

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
cioan: @mogoodrich , there are over 100 matches. are we sure we want to change all those strings?

Screenshot 2024-09-19 at 10 58 52 AM

mogoodrich: Fair enough, let's leave them as-is @cioan

PR #525, opened 2024-09-20
SL-667: Auto enrollment in the Pregnancy Program from a form

mogoodrich:
(set program outcome)


cioan (approved):
LGTM! Thanks, @mogoodrich !

mseaton (changes requested):
I don't really understand the point of this, nor do I know if it is a good idea.  I'll comment on the ticket.

mogoodrich:
Per @mseaton 's commnets, we decided not to make this changed.  I could have just closed this changeset, but instead pretty much reverted it and changed it o include in the validations a check to make sure that outcome is null.  Merging this small change in.

PR #526, opened 2024-09-26
SL-733: add patient location widget

cioan:
@mogoodrich and @mseaton , I got this patient location widget to display the status and location as we discussed yesterday:

Screenshot 2024-09-26 at 10 22 08 AM

Screenshot 2024-09-26 at 10 18 18 AM

Screenshot 2024-09-26 at 10 19 48 AM

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
@@ -1,0 +1,134 @@
+package org.openmrs.module.pihcore.fragment.controller.dashboardwidgets;
+
+
+import org.apache.commons.lang.StringUtils;
+import org.openmrs.Location;
+import org.openmrs.Patient;
+import org.openmrs.api.PatientService;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.appframework.domain.AppDescriptor;
+import org.openmrs.module.appui.UiSessionContext;
+import org.openmrs.module.emrapi.adt.AdtService;
+import org.openmrs.module.emrapi.adt.InpatientAdmission;
+import org.openmrs.module.emrapi.adt.InpatientAdmissionSearchCriteria;
+import org.openmrs.module.emrapi.adt.InpatientRequest;
+import org.openmrs.module.emrapi.adt.InpatientRequestSearchCriteria;
+import org.openmrs.module.emrapi.disposition.DispositionType;
+import org.openmrs.module.emrapi.patient.PatientDomainWrapper;
+import org.openmrs.module.emrapi.visit.VisitDomainWrapper;
+import org.openmrs.module.queue.api.QueueServicesWrapper;
+import org.openmrs.module.queue.api.search.QueueEntrySearchCriteria;
+import org.openmrs.module.queue.model.QueueEntry;
+import org.openmrs.ui.framework.UiUtils;
+import org.openmrs.ui.framework.annotation.FragmentParam;
+import org.openmrs.ui.framework.annotation.SpringBean;
+import org.openmrs.ui.framework.fragment.FragmentConfiguration;
+import org.openmrs.ui.framework.fragment.FragmentModel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class PatientLocationFragmentController {
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    public void controller(@SpringBean("patientService") PatientService patientService,
+                           @SpringBean("adtService") AdtService adtService,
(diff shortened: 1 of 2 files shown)


mseaton (approved):
Code generally looks correct to me @cioan , nice work!  Not sure if the layout is exactly how  Fiona wanted it, but defer to your conversations with her - can you check?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
```diff
@@ -0,0 +1,134 @@
...
+        Location currentInpatientLocation = null;
+        if (activeVisit != null ) {
+            // #1 Check if patient was admitted
+            InpatientAdmissionSearchCriteria criteria = new InpatientAdmissionSearchCriteria();
+            criteria.setPatientIds(Arrays.asList(patient.getPatientId()));
+            criteria.setIncludeDischarged(false);
+            List<InpatientAdmission> inpatientAdmissions = adtService.getInpatientAdmissions(criteria);
+            InpatientRequest currentInpatientRequest = null;
+            if (inpatientAdmissions != null && !inpatientAdmissions.isEmpty()) {
+                for (InpatientAdmission inpatientAdmission : inpatientAdmissions) {
```
mogoodrich, line 78: I think the assumption here is that there should only ever be one inpatient admission per patient, but, of course the query returns a list since it's a search?  If I'm understanding it, I think it would be better just to use inpatientAdmissions.get(0), but log a warning if inpatientAdmissins.size() > 2?

Or you could keep it this way but just add a comment saying there should never be more than one?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
```diff
@@ -0,0 +1,134 @@
...
+                    currentInpatientLocation = inpatientAdmission.getCurrentInpatientLocation();
+                    currentInpatientRequest = inpatientAdmission.getCurrentInpatientRequest();
+                    if (currentInpatientRequest != null) {
+                        String dispositionLocation = currentInpatientRequest.getDispositionLocation().getName();
+                        if (currentInpatientRequest.getDispositionType() == DispositionType.ADMIT) {
+                            patientStatus = ui.message("pihcore.waiting.tobe.admitted") + " " + dispositionLocation;
+                        } else if (currentInpatientRequest.getDispositionType() == DispositionType.TRANSFER) {
+                            patientStatus = ui.message("pihcore.waiting.for.transfer") + " " + dispositionLocation;
+                        } else {
+                            patientStatus = currentInpatientRequest.getDisposition().getName().getName();
```
mogoodrich, line 88: We might want to just display "Admitted" here if we don't find a status we recognize?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
```diff
@@ -0,0 +1,134 @@
...
+                List<InpatientRequest> inpatientRequests = adtService.getInpatientRequests(requestSearchCriteria);
+                if (inpatientRequests != null && !inpatientRequests.isEmpty()) {
+                    for (InpatientRequest inpatientRequest : inpatientRequests) {
+                        currentInpatientLocation = inpatientRequest.getDispositionLocation();
+                        if (inpatientRequest.getDispositionType() == DispositionType.ADMIT) {
+                            patientStatus = ui.message("pihcore.waiting.tobe.admitted") + " " + currentInpatientLocation;
+                        } else if (inpatientRequest.getDispositionType() == DispositionType.TRANSFER) {
+                            patientStatus = ui.message("pihcore.waiting.for.transfer") + " " + currentInpatientLocation;
+                        } else {
+                            patientStatus = inpatientRequest.getDisposition().getName().getName();
```
mogoodrich, line 108: Ditto here

mogoodrich (approved):
A couple comments, but generally LGTM @cioan , thanks!

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
cioan: @mogoodrich , I think this could be DISCHARGE which it should not happen in our system yet. But we could support this later? Maybe, I should return DISCHARGE then?

cioan:
Thanks!

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PatientLocationFragmentController.java
mogoodrich: Probably fine to keep as-is and just keep an eye on what might show up in it...

PR #527, opened 2024-09-27
SL-735: add Risk Factors dashboard widget

cioan:
I was able to add this widget to the dashboard with the following config:

Screenshot 2024-09-27 at 3 40 47 PM

Screenshot 2024-09-27 at 3 39 10 PM


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/RiskFactorsFragmentController.java
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        if (patient == null) {
+            throw new IllegalArgumentException("No patient found. Please pass patient into the configuration");
+        }
+
+        model.put("app", app);
+        model.put("patient", patient);
+
+        JsonNode node = app.getConfig().get("concepts");
```
mseaton, line 68: I'm a little confused about this configuration.  I would expect to see this kind of configuration if we named this fragment something more generic (eg. ObsByEncounterFragmentController), which we could use for various use cases, one of which is to display Risk Factors.  But I don't really understand the configuration if we are naming this controller so specifically.  Can we rename these to something more generic please?  Or if these are hard-coded in a way that is only appropriate for risk factors, can we get rid of the configuration?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/RiskFactorsFragmentController.java
```diff
@@ -0,0 +1,136 @@
...
+                patient = ((PatientDomainWrapper) patientConfig).getPatient();
+            } else if (patientConfig instanceof Integer) {
+                patient = patientService.getPatient((Integer) patientConfig);
+            } else if (patientConfig instanceof String) {
+                patient = patientService.getPatientByUuid((String) patientConfig);
+            }
+        }
+        if (patient == null) {
+            throw new IllegalArgumentException("No patient found. Please pass patient into the configuration");
+        }
```
mseaton, line 63: It would be nice if we didn't have all of this mess supporting different ways of loading the patient.  I know this is in a lot of existing coreapps controllers, but this is brand new, and something only we are including.  Presumably we know exactly how the patient will be pass in all the time, and we can assume that?

mseaton (changes requested):

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/RiskFactorsFragmentController.java
mogoodrich: Yeah, agree it makes sense to simplify that.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/RiskFactorsFragmentController.java
mogoodrich: Yeah, this looks like just a basic Obs-In-Program widget, and we could just call it something like that, but @cioan let me know if we are missing something.

cioan:
Thanks @mogoodrich and @mseaton for your reviews and comments! I have updated the PR to reflect your suggestions. Thanks!

omod/src/main/webapp/fragments/dashboardwidgets/obsByEncounter.gsp
```diff
@@ -0,0 +1,61 @@
...
+            <i class="icon-share-alt edit-action" title="Edit"></i>
+        </a>
+    </div>
+    <% if ( encounterObs !=null && (encounterObs.size() > 0 ))  { %>
+        <div class="info-body">
+            <table id="obsTable">
+                <thead>
+                <tr>
+                    <th>${ ui.message("coreapps.date") } </th>
+                    <th>${ ui.message("pihcore.riskFactors") }</th>
```
mseaton, line 35: @cioan - this is not correct anymore, now that this is generic.  This needs to probably be something passed in via config.  Can you fix this?

mogoodrich: Nice catch @mseaton , I missed this one...

PR #529, opened 2024-09-30
SL-735: add headers column config parameter

cioan:
@mseaton and @mogoodrich , here is the new "headers" config parameter for this widget. Thanks!

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/ObsByEncounterFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/ObsByEncounterFragmentController.java
@@ -80,6 +80,8 @@ public class ObsByEncounterFragmentController {
             String header = i.next().getTextValue();
             if (StringUtils.isNotBlank(header)) {
                 headers.add(header);
+            } else {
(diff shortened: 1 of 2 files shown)


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/ObsByEncounterFragmentController.java
```diff
@@ -70,6 +70,19 @@ public void controller(@SpringBean("patientService") PatientService patientServi
...
         }
+        node = app.getConfig().get("headers");
+        if (node == null) {
+            throw new IllegalStateException("Missing configuration headers on widget");
+        }
+        List<String> headers = new ArrayList<>();
+        i = node.getElements();
+        while (i.hasNext()) {
+            String header = i.next().getTextValue();
+            if (StringUtils.isNotBlank(header)) {
```
mseaton, line 81: Not sure about this use case?  When should it allowed to be blank?  I'd think this should either fail hard, or provide a default.

cioan: Thanks @mseaton ! I have addressed your observation.

PR #531, opened 2024-10-10
SL-740: add Apgar scores widget

cioan:
Potentially, this could be used to display any discrete obs values using this flexible configuration:
Screenshot 2024-10-10 at 9 39 30 AM

Screenshot 2024-10-10 at 9 09 55 AM

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
@@ -34,7 +34,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-public class ApgarScoresFragmentController {
+public class DiscreteValuesFragmentController {
 
     private final Logger log = LoggerFactory.getLogger(getClass());
 
@@ -130,7 +130,6 @@ public class ApgarScoresFragmentController {
                 simpleObject.put("obsValue", obs.getValueAsString(Context.getLocale()));
             }
         }
-        model.put("visitUrl", visitUrl);
         model.put("fields", fields);
     }
 
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
@@ -1,14 +1,9 @@
 package org.openmrs.module.pihcore.fragment.controller.dashboardwidgets;
 
 import org.apache.commons.lang.StringUtils;
-import org.codehaus.jackson.JsonNode;
-import org.codehaus.jackson.node.ArrayNode;
 import org.openmrs.Concept;
-import org.openmrs.Encounter;
 import org.openmrs.Obs;
 import org.openmrs.Patient;
-import org.openmrs.PatientProgram;
-import org.openmrs.Program;
 import org.openmrs.api.ConceptService;
 import org.openmrs.api.ObsService;
 import org.openmrs.api.ProgramWorkflowService;
@@ -24,17 +19,40 @@ import org.openmrs.ui.framework.fragment.FragmentModel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+/**
+ * This widget could be used to display discret obs values across many encounters within the same program enrollment.
+ * The config parameters are an array of fields, where each field has a:
+ *  - label : could indicate the name of the value to be displayed
+ *  - concept: points to the Concept of the obs whose value is displayed
+ *  - minValue: if specified and the obs.valueNumeric < minValue then the value will be highlighted in red
+ *
+ *      "fragmentConfig": {
+ *             "fields": [
+ *               {
+ *                 "label" : "pihcore.apgar.tenMinutes",
+ *                 "concept": "${concept.apgarAtTenMinutes.uuid}",
+ *                 "minValue": "7"
+ *               },
+ *               {
+ *                 "label" : "pihcore.apgar.fiveMinutes",
+ *                 "concept": "${concept.apgarAtFiveMinutes.uuid}",
+ *                 "minValue": "7"
+ *               },
+ *               {
+ *                 "label" : "pihcore.apgar.oneMinute",
+ *                 "concept": "${concept.apgarAtOneMinute.uuid}",
+ *                 "minValue": "7"
+ *               }
+ *             ],
+ *             "duringCurrentEnrollmentInProgram": "${program.infant.uuid}"
+ *           }
+ */
 public class DiscreteValuesFragmentController {
 
     private final Logger log = LoggerFactory.getLogger(getClass());
@@ -71,15 +89,23 @@ public class DiscreteValuesFragmentController {
                 String label = (String) fieldNode.get("label");
                 String conceptUuid = (String) fieldNode.get("concept");
                 String minValue = (String) fieldNode.get("minValue");
+                float floatValue = 0;
+                if (StringUtils.isNotBlank(minValue)) {
+                    try {
+                        floatValue = Float.parseFloat(minValue);
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("minValue config parameter has to be number", e);
+                    }
+                }
                 if (StringUtils.isNotBlank(conceptUuid)) {
                     Concept concept = conceptService.getConceptByReference(conceptUuid);
                     if (concept == null) {
-                        throw new IllegalArgumentException("No concept with this UUID: " + conceptUuid + " found. Please pass correct concept UUID into the configuration");
+                        throw new IllegalArgumentException("No concept with this ref: " + conceptUuid + " found. Please pass correct concept ref into the configuration");
                     }
                     concepts.add(concept);
                     fields.put(conceptUuid, SimpleObject.create(
                             "label", label,
-                            "minValue", minValue
+                            "minValue", floatValue
                     ));
                 }
             } catch (Exception e) {
@@ -94,7 +120,7 @@ public class DiscreteValuesFragmentController {
             obs.getValueAsString(Context.getLocale());
             SimpleObject simpleObject = fields.get(obs.getConcept().getUuid());
             if (simpleObject != null ) {
-                simpleObject.put("obsValue", obs.getValueAsString(Context.getLocale()));
+                simpleObject.put("obs", obs);
             }
         }
         model.put("fields", fields);
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
@@ -15,6 +15,7 @@ import org.openmrs.api.ProgramWorkflowService;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.appframework.domain.AppDescriptor;
 import org.openmrs.module.emrapi.patient.PatientDomainWrapper;
+import org.openmrs.module.pihcore.PihCoreUtils;
 import org.openmrs.ui.framework.SimpleObject;
 import org.openmrs.ui.framework.annotation.FragmentParam;
 import org.openmrs.ui.framework.annotation.SpringBean;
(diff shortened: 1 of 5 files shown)


mseaton (changes requested):
Per your comment @cioan , we have the exact same issue here that happened in the last widget - you wrote a useful, generic fragment, but named it for a specific use case, and did not pull out the one or two hard-coded things, namely the label.  Can you please rename this and pull out anything hard-coded to "apgar" into a config property, just like before?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,143 @@
...
+            patient = ((PatientDomainWrapper) patientConfig).getPatient();
+        }
+        if (patient == null) {
+            throw new IllegalArgumentException("No patient found. Please pass patient into the configuration");
+        }
+
+        model.put("app", app);
+        model.put("patient", patient);
+
+        JsonNode node = app.getConfig().get("fields");
```
mseaton, line 61: I think this same issue probably exists everywhere / all over the place, but this is something I was contending with in other fragments I was recently writing for Rwanda.  Basically, I think all of the config that drives the behavior of the fragment should be pulled from the `FragmentConfiguration config`, _not_ from the `app.getConfig()`.  The app config is for configuring the "app", whereas the fragment config should be for configuring the fragment.

Think of it this way - if you wanted to add this same fragment to a gsp directly (eg. without adding it as an app/extension), could you do it?  If the configuration it needs is coming from an "app.getConfig()" then you couldn't.  But if it is coming from the fragment config, then you simply put the configuration into the fragment declaration on the page, and it works as expected.

Does this make sense?  The same applies to any configurations here

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,143 @@
...
+        if (node instanceof ArrayNode) {
+            ArrayNode arrayNode = (ArrayNode) node;
+            for (Iterator<JsonNode> iter = arrayNode.getElements(); iter.hasNext();) {
+                JsonNode fieldNode = iter.next();
+                try {
+                    String label = fieldNode.get("label").asText();
+                    String conceptUuid = fieldNode.get("concept").asText();
+                    String minValue = fieldNode.get("minValue").asText();
+                    if (StringUtils.isNotBlank(conceptUuid)) {
+                        Concept concept = conceptService.getConceptByUuid(conceptUuid);
```
mseaton, line 76: Ideally this would support mappings equally well.  You could use `HtmlFormEntryUtil.getConcept(String lookup)` or the new method in core `conceptService.getConceptByReference(String ref)`, which supports uuid and a number of other lookup options.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,143 @@
...
+                    throw new IllegalStateException("Invalid configuration for encounterType node", e);
+                }
+            }
+        } else {
+            throw new IllegalStateException("fields is expected to be an array of objects");
+        }
+
+        Date obsOnOrAfter = null;
+        String visitUrl = getConfigValue(app, "visitUrl");
+        String programUuid = getConfigValue(app, "duringCurrentEnrollmentInProgram");
```
mseaton, line 97: At this point, it is probably high time that we move some of what we are doing here into a common set of utility methods.  We now have a number of controllers that all have similar logic for getting encounter and obs that fall within the current enrollment of a particular program.  And we will likely continue to need more of this.  Can we look at extracting this to some utilities that we can re-use, test, and ensure bug fixes are applied universally?

mseaton (changes requested):
See my new comments @cioan

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
cioan: @mseaton , makes sense. How would you do this? Is this fragment config configured this way anywhere else? I could not find an example on how to change it. Thanks!

cioan: @mseaton , is this what you meant? using fragmentConfig param?
Screenshot 2024-10-10 at 2 08 10 PM

cioan: fyi @mseaton , the above, using the fragmentConfig worked fine. I will refactor this widget to use the fragmentConfig node.

mseaton: Yes, that's right @cioan

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Arrays;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
```
mseaton, line 37: It is not at all obvious what this fragment does from the code below.  Can you add a comment describing what this does and when it would be used and how you'd configure it?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,102 @@
...
+            throw new IllegalStateException("Missing configuration fields on widget");
+        }
+        Set<Concept> concepts = new HashSet<>();
+        Map<String, SimpleObject> fields = new LinkedHashMap<>();
+        for (int i = 0; i < fieldsObject.size(); i++ ) {
+            try {
+                Map<String, String> fieldNode = (LinkedHashMap) fieldsObject.get(i);
+                String label = (String) fieldNode.get("label");
+                String conceptUuid = (String) fieldNode.get("concept");
+                String minValue = (String) fieldNode.get("minValue");
```
mseaton, line 73: I have to imagine that maxValue may also be a thing we need to support, though I guess that can come later as needed

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        String programUuid = (String) config.get("duringCurrentEnrollmentInProgram");
+        List<Obs> obsList = PihCoreUtils.getObsWithinProgram(patient, concepts, programUuid);
+
+        for (Obs obs : obsList) {
+            obs.getValueAsString(Context.getLocale());
+            SimpleObject simpleObject = fields.get(obs.getConcept().getUuid());
+            if (simpleObject != null ) {
+                simpleObject.put("obsValue", obs.getValueAsString(Context.getLocale()));
```
mseaton, line 97: Why are we getting value as String here?  Why not just put the entire obs in here?  It's easy enough to format this as a String in the gsp by using `ui.format(obs)` (I think), and don't you need to get the numeric value for usage in the gsp to determine if it is outside of range?

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
```diff
@@ -0,0 +1,23 @@
...
+        <a href="${ ui.urlBind("/" + contextPath + app.url, [ "patient.uuid": patient.id ]) }" class="right">
+            <i class="icon-share-alt edit-action" title="Edit"></i>
+        </a>
+    </div>
+    <div class="info-body">
+        <% fields.each { concept, obs -> %>
+            <div>
+                <span style="font-family: 'OpenSansBold'";>${ ui.message(obs.label) }:</span>
+                <% if ( obs.minValue && obs.obsValue && obs.minValue.isNumber() && obs.obsValue.isNumber() &&
+                        (new BigDecimal(obs.obsValue) < new BigDecimal(obs.minValue))) {%>
```
mseaton, line 15: Per my comment above, shouldn't we just put the obs in the object, and then do a comparison on obs.valueNumeric?  What is this `isNumber()` method you are calling?  Is that some magic groovy thing?  Does that work?

Basically if `obs.valueNumeric && ...`

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
```diff
@@ -0,0 +1,23 @@
...
+                    <span style="color: red;">${ obs.obsValue }</span>
```
mseaton, line 16: Would be best to pass a css class in and/or default to one here.  We can then style that in our global.css file.  This is fine for now I guess, but we may want to think about that for the future.

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
```diff
@@ -0,0 +1,23 @@
...
+        <h3>${ ui.message(app.label) }</h3>
+        <a href="${ ui.urlBind("/" + contextPath + app.url, [ "patient.uuid": patient.id ]) }" class="right">
+            <i class="icon-share-alt edit-action" title="Edit"></i>
+        </a>
+    </div>
+    <div class="info-body">
+        <% fields.each { concept, obs -> %>
+            <div>
+                <span style="font-family: 'OpenSansBold'";>${ ui.message(obs.label) }:</span>
+                <% if ( obs.minValue && obs.obsValue && obs.minValue.isNumber() && obs.obsValue.isNumber() &&
```
mseaton, line 14: We should add `minValue` as a number the first place, and validate that it's a number in the Controller.  By the time we  get here, minValue should be assumed to either be null or a number.

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
```diff
@@ -0,0 +1,23 @@
+
+<div class="info-section">
+    <div class="info-header">
+        <i class="${app.icon}"></i>
+        <h3>${ ui.message(app.label) }</h3>
+        <a href="${ ui.urlBind("/" + contextPath + app.url, [ "patient.uuid": patient.id ]) }" class="right">
+            <i class="icon-share-alt edit-action" title="Edit"></i>
```
mseaton, line 7: Is this an "Edit"?  If not, it shouldn't be listed that way.  And...what is this exactly?  What would one actually be linking to or editing in such a widget that spans multiple obs across encounters within a program?  Is this in the requirements or is this just copied from some other fragment and you forgot to remove it?

mseaton (changes requested):
Looking better @cioan .  But now I'm actually looking more into the details rather than the bigger structure.  See a few additional comments and let me know if you want to discuss.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
```diff
@@ -0,0 +1,102 @@
...
+        for (int i = 0; i < fieldsObject.size(); i++ ) {
+            try {
+                Map<String, String> fieldNode = (LinkedHashMap) fieldsObject.get(i);
+                String label = (String) fieldNode.get("label");
+                String conceptUuid = (String) fieldNode.get("concept");
+                String minValue = (String) fieldNode.get("minValue");
+                if (StringUtils.isNotBlank(conceptUuid)) {
+                    Concept concept = conceptService.getConceptByReference(conceptUuid);
+                    if (concept == null) {
+                        throw new IllegalArgumentException("No concept with this UUID: " + conceptUuid + " found. Please pass correct concept UUID into the configuration");
```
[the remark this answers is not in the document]
cioan: ok

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
cioan: added

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
cioan: yes, we could add this later if needed

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/DiscreteValuesFragmentController.java
cioan: ok

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
cioan: ok

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
cioan: ok, we will think about this in the future

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
cioan: ok

omod/src/main/webapp/fragments/dashboardwidgets/discreteValues.gsp
cioan: deleted

cioan:
@mseaton , I think that I addressed again all your latest suggestions. Thanks!

mseaton (approved):
Thanks for bearing with me over multiple reviews!

PR #533, opened 2024-10-16
SL-742: implement Newborn Indicators widget

cioan:
Since the Newborn Indicators widget is required to display the patient location, I have moved the code from the Patient Location widget into a common class that both widgets could invoke. 

Screenshot 2024-10-16 at 10 07 04 AM


mseaton (approved):
Thanks @cioan .  Looks ok to me at first glance.  I might have initially expected us to build a widget like this with the status data functionality that we have, but this is fine and likely easier.

cioan:
Thanks @mseaton !

PR #536, opened 2024-10-22
SL-786: Set Outcome to "Treatment Completed" when Treatment Status is…

mogoodrich:
… "Miscarried"

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/task/ClosePregnancyProgramTask.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/task/ClosePregnancyProgramTask.java
@@ -18,6 +18,7 @@ import java.util.List;
 /**
  * Patients should be unenrolled automatically from the pregnancy program if they have been in the Antenatal Treatment State for 11 months. The program outcome should be Lost to Follow Up.
  * Patients should be unenrolled automatically from the pregnancy program 6 weeks after the Delivered treatment state start date with an outcome of Treatment Completed
+ * For any programs with a completion date within the last year, if the final state is Miscarried and there is no outcome, the outcome should be set to Treatment Completed
  */
 public class ClosePregnancyProgramTask implements Runnable {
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/pihcore/task/PihCoreScheduledTaskExecutor.java
```diff
@@ -23,14 +21,15 @@ public class PihCoreScheduledTaskExecutor extends ScheduledExecutorFactoryBean {
...
+    private final long twentyfourHours = oneHour * 24;
 
     public PihCoreScheduledTaskExecutor() {
         setScheduledExecutorTasks(
                 task(fiveMinutes, fourHours, MarkAppointmentSchedulingAppointmentsAsMissedOrCompletedTask.class),
                 task(fiveMinutes, oneHour, PihCloseStalePullRequestsTask.class),
                 task(fiveMinutes, oneHour, PihCloseStaleCreateRequestsTask.class),
                 task(fiveMinutes, oneHour, PihCloseStaleVisitsTask.class),
-                task(fiveMinutes, oneHour, ClosePregnancyProgramTask.class),
+                task(fiveMinutes, twentyfourHours, ClosePregnancyProgramTask.class),
```
mogoodrich, line 32: It occurred to me that there is no need (besides to ease testing :) to run this task more than once per day.

mseaton: I was going to make the same comment, but then I figured we must do this for a reason, as the same is true for most/all of the other tasks.  In particular, if the system is not reliably up 24/7, I'd worry that the task might not execute every day if the server was down at the given execution time each day.

api/src/main/java/org/openmrs/module/pihcore/task/ClosePregnancyProgramTask.java
```diff
@@ -62,7 +64,21 @@ public void run() {
...
+                    }
+                }
+
+                // now fetch all pregnancy programs that *have* been completed within the last year, and if their final state is miscarried, set the outcome to treatment complete
+                // note: when back entering information from more than a year ago, this will not work as expected, as for performance reasons it will only look at the last year
+                patientPregnancyPrograms = Context.getProgramWorkflowService().getPatientPrograms(null, pregnancyProgram, null, null, new DateTime().minusYears(1).toDate(), new Date(), true);
+                for (PatientProgram patientPregnancyProgram : patientPregnancyPrograms) {
+                    if (patientPregnancyProgram.getDateCompleted() != null && patientPregnancyProgram.getOutcome() == null) {
+                        for (PatientState patientState : patientPregnancyProgram.getMostRecentStateInEachWorkflow()) {
+                            if (patientState.getState().equals(miscarriedState)) {
```
mseaton, line 80: Probably don't need to check workflow here, but in case we do, figured I'd mention it

mseaton (approved):
Looks good, but see a few comments I made and address as appropriate.

api/src/main/java/org/openmrs/module/pihcore/task/PihCoreScheduledTaskExecutor.java
mogoodrich: The other tasks are more time sensitive, as they are all on the order of hours (close a visit after 12 hours with no encounters vs close a program after 6 weeks in a certain state).  And all these tasks are set to run for the first time 5 minutes after startup, so we shouldn't need to worry about the something not running because the server is down?  @mseaton   Easy enough to move it to 4 or 12 hours if that feels safer, though.

mogoodrich: I think I'll change this to twelve hours...

api/src/main/java/org/openmrs/module/pihcore/task/ClosePregnancyProgramTask.java
mogoodrich: Yep, nope... only one workflow for this program (and assumedly if there were more, miscarried would only be in one of them)

PR #541, opened 2024-11-19
SL-731: add pregnancy indicators widget

cioan:
Here is the new Pregnancy Indicators widget, as it was mocked on the SL-731 ticket:
Screenshot 2024-11-19 at 10 30 03 AM

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
@@ -64,7 +64,7 @@ public class PregnancyIndicatorsFragmentController {
                 "css", "red",
                 "obs", ui.message("pihcore.pregnant")
         ));
-        Concept question = conceptService.getConceptByUuid(PihEmrConfigConstants.CONCEPT_ESTIMATEDDUEDATE_UUID);
+        Concept question = conceptService.getConceptByUuid(SierraLeoneConfigConstants.CONCEPT_ESTIMATEDDUEDATE_UUID);
         List<Obs> obsWithinProgram = null;
         if ( question != null ) {
             obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, programUuid);
@@ -86,21 +86,48 @@ public class PregnancyIndicatorsFragmentController {
                 "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? ui.message("general.yes") : ui.message("general.no")
         ));
 
-        question = conceptService.getConceptByUuid(SierraLeoneConfigConstants.CONCEPT_DATETIMEOFBIRTH_UUID);
+        question = conceptService.getConceptByUuid(PihEmrConfigConstants.CONCEPT_DELIVERYDATEANDTIME_UUID);
+        Date latestDeliveryDate = null;
+        Obs latestDeliveryObs = null;
         if ( question != null ) {
             obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, null);
+            for (Obs obs : obsWithinProgram) {
+                Date obsValueDate = obs.getValueDate();
+                if ( (obsValueDate!= null)  &&
+                        (latestDeliveryDate == null || latestDeliveryDate.before(obs.getValueDate()))) {
+                    latestDeliveryDate = obs.getValueDate();
+                    latestDeliveryObs = obs;
+                }
+            }
         }
         fields.put("latestDelivery", SimpleObject.create(
                 "label", "pihcore.latest.delivery",
-                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? obsWithinProgram.get(0) : ""
+                "obs", ( latestDeliveryDate !=null ) ? latestDeliveryDate : ""
         ));
-        question = conceptService.getConceptByUuid(PihEmrConfigConstants.CONCEPT_TYPEOFDELIVERY_UUID);
-        if ( question != null ) {
-            obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, null);
+
+        Set<Obs> methodsOfDelivery = new HashSet<>();
+        if ( latestDeliveryObs != null ) {
+            Encounter encounter = latestDeliveryObs.getEncounter();
+            // search for all the Methods of Delivery associated with the encounter (form) that contains the most recent date/delivery time
+            for (Obs obs : encounter.getAllObs(false) ) {
+                if (obs.getConcept().getUuid().equals(PihEmrConfigConstants.CONCEPT_TYPEOFDELIVERY_UUID)) {
+                    // check if this method of delivery was already added to the list
+                    boolean found = false;
+                    for (Obs method: methodsOfDelivery) {
+                        if ( method.getValueCoded().getUuid().equalsIgnoreCase(obs.getValueCoded().getUuid()) ) {
+                            found = true;
+                        }
+                    }
+                    if (!found) {
+                        methodsOfDelivery.add(obs);
+                    }
+                }
+            }
         }
+
         fields.put("deliveryType", SimpleObject.create(
                 "label", "pihcore.latest.delivery.type",
-                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? obsWithinProgram.get(0) : ""
+                "obs", (methodsOfDelivery != null && methodsOfDelivery.size() > 0 ) ? methodsOfDelivery : ""
         ));
 
         EncounterSearchCriteriaBuilder searchCriteriaBuilder = new EncounterSearchCriteriaBuilder();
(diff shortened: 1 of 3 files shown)


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
```diff
@@ -0,0 +1,157 @@
...
+                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? ui.message("general.yes") : ui.message("general.no")
+        ));
+
+        question = conceptService.getConceptByUuid(SierraLeoneConfigConstants.CONCEPT_DATETIMEOFBIRTH_UUID);
+        if ( question != null ) {
+            obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, null);
+        }
+        fields.put("latestDelivery", SimpleObject.create(
+                "label", "pihcore.latest.delivery",
+                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? obsWithinProgram.get(0) : ""
```
mogoodrich, line 95: Definitely an edge case for twins born a significant time apart, but if there are multiple obs found, shouldn't we take the latest delivery date.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
```diff
@@ -0,0 +1,157 @@
...
+        ));
+        question = conceptService.getConceptByUuid(PihEmrConfigConstants.CONCEPT_TYPEOFDELIVERY_UUID);
+        if ( question != null ) {
+            obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, null);
+        }
+        fields.put("deliveryType", SimpleObject.create(
+                "label", "pihcore.latest.delivery.type",
+                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? obsWithinProgram.get(0) : ""
```
mogoodrich, line 103: See my comment above.  Those this is a little trickier to figure out since you'd likely need to get the parent obs group. Maybe if there are more than one obs and they are different, just display both of them?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
```diff
@@ -0,0 +1,157 @@
...
+        ancEncTypes.add(encounterService.getEncounterTypeByUuid(PihEmrConfigConstants.ENCOUNTERTYPE_ANC_FOLLOWUP_UUID));
+
+        searchCriteriaBuilder = new EncounterSearchCriteriaBuilder();
+        searchCriteriaBuilder.setPatient(patient);
+        searchCriteriaBuilder.setEncounterTypes(ancEncTypes);
+        if ( programUuid != null ) {
+            searchCriteriaBuilder.setFromDate(PihCoreUtils.getProgramEnrollmentDate( patient, programUuid ));
+        }
+        encounters = encounterService.getEncounters(searchCriteriaBuilder.createEncounterSearchCriteria());
+        int numberOfVisits = encounters != null ? encounters.size() : 0;
```
mogoodrich, line 134: Looking at the specifications, this follows from that, but wouldn't we want to collapse this by visits, ie the number of visits with an ANC Intake or ANC Followup encounter?  Granted, a visit should only have one, but this would prevent against over-counting in the case of duplicate form entry?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
```diff
@@ -0,0 +1,157 @@
...
+        for (Encounter encounter : encounters) {
+            if (encounter.getEncounterType().getUuid().equals(PihEmrConfigConstants.ENCOUNTERTYPE_ANC_INTAKE_UUID)) {
+                for (Obs candidate : encounter.getObsAtTopLevel(false)) {
+                    if (candidate.getConcept().getUuid().equalsIgnoreCase(SierraLeoneConfigConstants.CONCEPT_NUMBEROFANCVISITS_UUID)) {
+                        Double valueNumeric = candidate.getValueNumeric();
+                        if (!valueNumeric.isNaN()) {
+                            // the question on the form is "What number ANC Visit is this?", therefore it should include the ANC Intake visit
+                            numberOfVisits = numberOfVisits + valueNumeric.intValue() - 1;
```
mogoodrich, line 142: I understand the logic here but will be interesting to see how this works out in practice.  The idea is that ideally this will always be "1" (the Intake is the first visit?) but this is to capture the case where somehow there were visits prior to Intake that were not recorded in the EMR? @fiona855

mogoodrich (approved):
Generally looks good, thanks @cioan , a couple comments/suggestions.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
cioan: Yes, the method we are using here https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/java/org/openmrs/module/pihcore/PihCoreUtils.java#L53 

actually orders the Obs descending by obsDatetime:
https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/db/hibernate/HibernateObsDAO.java#L275

But there is a possibility that someone enter manually the second baby born at a sooner date than the first baby? I could edit the code to specifically look at the value of the obs and get the most recent obs.value?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
cioan: Sorry, I did not understand your question here. We discussed this yesterday during the standup, and we could discuss more tomorrow?

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
cioan: Oh, now I realized that you missed our discussion from yesterday, @mogoodrich ?

mogoodrich: Oh, was there a discussion at standup that I missed? No worries, I don't mean to relitigate if you all have already discussed it.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
```diff
@@ -0,0 +1,157 @@
...
+        List<Obs> obsWithinProgram = null;
+        if ( question != null ) {
+            obsWithinProgram = PihCoreUtils.getObsWithinProgram(patient, new HashSet<>(Arrays.asList(question)), null, programUuid);
+        }
+
+        fields.put("dueDate", SimpleObject.create(
+                "label", "pihcore.delivery.dueDate",
+                "css", "red",
+                "obs", (obsWithinProgram!=null && obsWithinProgram.size() > 0 ) ? obsWithinProgram.get(0) : ""
+        ));
```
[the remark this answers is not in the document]
mogoodrich: A fair point, agreed.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
cioan: @mogoodrich , I have updated the PR to display all different methods of delivery that are on the same Labour and Delivery form as the most recent Delivery date/time obs value:

Screenshot 2024-11-22 at 6 26 02 AM

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/PregnancyIndicatorsFragmentController.java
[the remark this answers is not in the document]
mogoodrich: another fair point, happy to discuss @cioan  (though I think we will be having a shortened standup, so might not make it today).  One intermediate option would be to create some sort of Sierra Leone status utility methods?   The key is really if we create another SL dashboard we don't duplicate the logic but reuse it to make sure it stays consistent across Sierra Leone implementations.fyi @mseaton

cioan:
Thanks @mogoodrich ! Since now we are in the agreement on how to calculate those indicators, I will merge in this PR. Next week we will discuss more how to refactor this code, and will create new PR. But  I just wanted to give Fiona a chance to validate the calculations of these indicators and the widget UI which it should not change when we refactor the code.

mogoodrich:
Sounds good (looks like the build broke though)?

cioan:
@mogoodrich , the pihcore built before the config-pisl which contained the newly defined concept uuids. It is building successfully now. Have a good weekend!

mogoodrich:
> @mogoodrich , the pihcore built before the config-pisl which contained the newly defined concept uuids. It is building successfully now. Have a good weekend!

Thanks! You too!

PR #542, opened 2024-11-21
(chore) UHM-8241 - add html classes to various widgets

chibongho:
Add html class names to various widgets. This makes is easier to select and hide them when embedding the O2 pregnancy dashboard into an O3 workspace.

image

For widgets that are configurable to show different data (like the encounters widget), I added `${app.id}` (ex: "pregnancy.dashboard.ancInitialEncounters") as a class name. For widgets that are not configurable to show different data (like the appointments widget) I just hard-coded a string.

I'll make a corresponding PR for widgets in `opnmrs-module-coreapps` as well.

### DIFF
--- a/omod/src/main/webapp/fragments/dashboardwidgets/newbornIndicators.gsp
+++ b/omod/src/main/webapp/fragments/dashboardwidgets/newbornIndicators.gsp
@@ -3,7 +3,7 @@
         color: red;
     }
 </style>
-<div class="info-section">
+<div class="info-section newborn-indicator">
     <div class="info-header">
         <i class="${app.icon}"></i>
         <h3>${ ui.message(app.label) }</h3>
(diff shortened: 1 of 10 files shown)


omod/src/main/webapp/fragments/dashboardwidgets/newbornIndicators.gsp
```diff
@@ -3,7 +3,7 @@
         color: red;
     }
 </style>
-<div class="info-section">
+<div class="info-section newborn-indicator">
```
cioan, line 6: Shouldn't this be ${app.id} similar to the other ones?

omod/src/main/webapp/fragments/dashboardwidgets/patientLocation.gsp
```diff
@@ -1,5 +1,5 @@
 <% if (patientStatus.length() > 0 ) { %>
-    <div class="info-section">
+    <div class="info-section patient-location">
```
cioan, line 2: ${app.id} here too?

cioan (approved):
@chibongho , I've had just a couple minor comments. Thanks!

mogoodrich (approved):

PR #551, opened 2025-01-06
UHM-8245 give unique names to twins/triplets created from labour/devl…

chibongho:
…ivery summary form

Note:
When testing this by submitting a "labour / delivery summary" form, the submission only partial succeeds (more detail at [UHM-8343](https://pihemr.atlassian.net/browse/UHM-8343)). That said, the babies are created properly with unique names ("baby", "baby 2", "baby 3").

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
@@ -75,7 +75,7 @@ public class RegisterBabyAction implements CustomFormSubmissionAction {
                                 Concept gender = getCodedValue(groupMembers, BABY_GENDER_CONCEPT_SOURCE, BABY_GENDER_CONCEPT);
                                 Date birthDatetime = getObsDateValue(groupMembers, DATE_TIME_OF_BIRTH_CONCEPT_SOURCE, DATE_TIME_OF_BIRTH_CONCEPT);
                                 if ((gender != null) && (birthDatetime != null)) {
-                                    Patient baby = registerBaby(mother, gender, birthDatetime, encounter.getLocation());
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -75,7 +75,7 @@ public void applyAction(FormEntrySession formEntrySession) {
                                 Concept gender = getCodedValue(groupMembers, BABY_GENDER_CONCEPT_SOURCE, BABY_GENDER_CONCEPT);
                                 Date birthDatetime = getObsDateValue(groupMembers, DATE_TIME_OF_BIRTH_CONCEPT_SOURCE, DATE_TIME_OF_BIRTH_CONCEPT);
                                 if ((gender != null) && (birthDatetime != null)) {
-                                    Patient baby = registerBaby(mother, gender, birthDatetime, encounter.getLocation());
+                                    Patient baby = registerBaby(mother, gender, birthDatetime, encounter.getLocation(), registeredBabies.size());
```
mogoodrich, line 78: I haven't looked at the rest of the code so not sure, but if this is within some sort of iterator don't we want the iterator count, not the total size?  ie it doesn't seem like current babyNumber = total number of babies?

mseaton: It seems like it does the right thing to me, as `registeredBabies` starts out empty and increases for each baby that is iterated over, so I think this should be fine @mogoodrich

mseaton (approved):

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
chibongho: Yeah, `registeredBabies` is only storing the already registered babies, so it starts out at size 0 before the loop and grows for each baby registered.

PR #556, opened 2025-02-06
UHM-8303 - Upgrade to OpenMRS 2.7.2

mseaton:


mogoodrich:
LGTM... let's just confirm with the team that there are no imminent releases planned (or at least time for testing beforehand) and the the build is (relatively) running before we merge KGH Test is down now and there was a deploy failure to Liberia but I'm hoping/thinking they are both transitory)

PR #568, opened 2025-03-21
HAI-1035 Added biometrics widget to the HIV dashboard

lnball:
@mogoodrich Can you confirm that this looks right?  I cannot test fully on my SDK because I don't have a fingerprint reader.  If it looks somewhat correct, please merge.  Then I can test on a server with fingerprints (ie. haitihivtest or humci)

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
@@ -136,6 +136,7 @@ public class CustomAppLoaderConstants {
         public static final String PATHOLOGY_SUMMARY = "labtrackingapp.labSummary";
         public static final String ID_CARD_PRINTING_STATUS = "mirebalais.idCardPrintingStatus";
         public static final String BIOMETRICS_SUMMARY = "registrationapp.biometricsSummary";
+        public static final String BIOMETRICS_WIDGET = "dashboard.biometricsSummary";
 
         public static final String RADIOLOGY_APP = "radiology.app";
         public static final String RADIOLOGY_ORDERS_APP = "radiology.orders.app";
@@ -594,7 +595,8 @@ public class CustomAppLoaderConstants {
             Apps.NOTES_SUMMARY,
             Apps.COVID_LAB_RESULTS,
             Apps.RECENT_LAB_RESULTS,
-            Apps.ALL_LAB_RESULTS
+            Apps.ALL_LAB_RESULTS,
+            Apps.BIOMETRICS_WIDGET
     );
 
     public static final List<String> CLINICIAN_DASHBOARD_SECOND_COLUMN_ORDER = Arrays.asList(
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
```diff
@@ -594,7 +595,8 @@ public static final class EncounterTemplates {
             Apps.NOTES_SUMMARY,
             Apps.COVID_LAB_RESULTS,
             Apps.RECENT_LAB_RESULTS,
-            Apps.ALL_LAB_RESULTS
+            Apps.ALL_LAB_RESULTS,
+            Apps.BIOMETRICS_WIDGET
```
mogoodrich, line 599: If we aren't adding this to the main clinicial dashboard we don't need to add this here.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -2869,11 +2867,14 @@ private void enableHIVProgram() {
                         "dashboard", programUuid   // provides contextual context so this widget knows which dashboard it's being rendered on
                 )),
                 "coreapps", "program/programHistory",
-                secondColumnIndex++
+                2
         ));
 
         // TODO: Add ACTIVE MEDICATIONS HERE
 
+        // Start index for order after fingerprints
```
mogoodrich, line 2875: That's annoying (that you can't just use the "++" because the biometrics widget is added elsewhere).

It took me a few minutes to figure out what was going on here, maybe make this comment "Start index for order after fingerprints (which are added in the biometrics section)"

mogoodrich (approved):
A couple comments but general LGTM @lnball

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderConstants.java
mogoodrich: Ah, but maybe this is a correction because it already is on the main clinician dashboard and we never provided an order for it?

lnball: Good point @mogoodrich .  The ticket description was to add to the main clinical dashboard but then it became clear that this was for the HIV dashboard only.  I will remove from this line.  Thanks.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
lnball: Right @mogoodrich .  I should have explained.  I don't think that we have implementations that don't have the HIV program AND Biometrics, but it's possible.  Figured it was cleaner to include the biometrics dashboard widget with the biometrics component but it makes for this messy ordering of the widgets.

I did test and if the biometrics widget is missing, it still orders widgets properly.

PR #569, opened 2025-04-03
HAI-1026 - Add opportunistic infections on the HIV Patient chart

mseaton:


omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/MostRecentObsWithCodedValueFragmentController.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The purpose of this is to render one row per distinct coded value for the given patient and the given configured
+ * `concept`, where the most recent observation per coded value is used.
+ * If a particular `codedValueSet` is configured, it limits the coded values rendered to just those within that set
+ */
+public class MostRecentObsWithCodedValueFragmentController {
```
mogoodrich, line 32: Unhelpful:  having "most recent" as the first part of this name is somewhat confusing, because this is really about finding *all* coded answered recorded for a particular concept, not the most recent.  (I get now that you mean the most recent obs for *each* coded value, but it took me a while to get there)

Maybe just "ObsCodedValuesFragmentController", and then, overall documentation is that for a given concept and patient find all coded answer values recorded for that concept and patient, displaying the obs datetime of the most recent recorded value of each obs value.

I tend to you think of "MostRecent" applying to numerical values that change over time.

mogoodrich (commented):
LGTM, maybe an unhelpful comment about documentation/naming?

PR #570, opened 2025-04-28
SL-884: calculate due date base on the gestational age

cioan:
@mogoodrich , just calculating the due date based on the gestational age and the obs datetime. Thanks!


mogoodrich (approved):
Generally LGTM, defer to @dmdesimone  to test the formula! :)

cioan:
Thanks @mogoodrich !

PR #571, opened 2025-05-02
UHM-8591: prevent NPE when patient has no birthdate

cioan:
@mogoodrich , here is the fix for the NPE when potential matches have no birthdate. I created a unit test and was able to reproduce the NPE we were seeing in humci yesterday before you cleaned up the test data. Thanks for reviewing!

### DIFF
--- a/api/src/test/java/org/openmrs/module/pihcore/search/PihPatientSearchAlgorithmTest.java
+++ b/api/src/test/java/org/openmrs/module/pihcore/search/PihPatientSearchAlgorithmTest.java
@@ -13,13 +13,17 @@ import org.openmrs.module.pihcore.PihCoreContextSensitiveTest;
 import org.openmrs.module.pihcore.metadata.Metadata;
 import org.openmrs.module.registrationcore.api.search.PatientAndMatchQuality;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.testcontainers.shaded.org.apache.commons.lang3.ObjectUtils;
 
 import java.util.Date;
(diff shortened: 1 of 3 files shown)


api/src/test/java/org/openmrs/module/pihcore/search/PihPatientSearchAlgorithmTest.java
```diff
@@ -13,13 +13,17 @@
 import org.openmrs.module.pihcore.metadata.Metadata;
 import org.openmrs.module.registrationcore.api.search.PatientAndMatchQuality;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.testcontainers.shaded.org.apache.commons.lang3.ObjectUtils;
```
mogoodrich, line 16: Not a blocker, but this is a weird library I haven't seen before?

mogoodrich (approved):
One aside, but LGTM, thanks @cioan !

api/src/test/java/org/openmrs/module/pihcore/search/PihPatientSearchAlgorithmTest.java
mseaton: Yeah, can you fix that @cioan ?  I assume it was just included inadvertently when you were going for a standard commons lang import?

cioan: Yes, thank you! I have removed that unused library. Thanks!

PR #572, opened 2025-05-05
UHM-8573: PIH Core Importer/Exporter: Fix bug when value numeric is 0

mogoodrich:


cioan:
Looks good! Thanks @mogoodrich !

PR #576, opened 2025-05-12
SL-931: add the current log on user as the provider of the registration encounter

cioan:
@mogoodrich , this is just adding the log on user, the user who fills in the labour and delivery form as the provider of the baby's registration encounter. Thanks!


mogoodrich (approved):
LGTM, thanks @cioan !

cioan:
Thanks @mogoodrich !

PR #577, opened 2025-05-16
SL-917: restrict one ANC intake per programent enrollment

cioan:


mogoodrich (commented):
If I'm following this correctly, I think this change stops the ANC Intake from being filled out if the patient is actively enrolled in the program? Is the assumption here is that the ANC Intake form is what is enrolled them in the program? But what if they were manually enrolled without filling out a form, would they never be able to fill out an Intake form that that program?

Also, does this properly support retrospectively adding an Intake form for a previous enrollment, or is it all based on whether the patient is enrolled in the program on the current date?

Not sure if all this is doable without a lot of work, but I think this what the requirements are asking for?

api/src/main/java/org/openmrs/module/pihcore/setup/AppFrameworkExpressionFunctionsSetup.java
```diff
@@ -0,0 +1,43 @@
...
+            if (scriptsDir.exists()) {
+                File[] functions = scriptsDir.listFiles();
+                if (functions != null && functions.length > 0) {
+                    for (File function : functions) {
+                        String functionName = function.getName();
+                        try (FileInputStream fileInputStream = new FileInputStream(function)) {
+                            String value = IOUtils.toString(Objects.requireNonNull(fileInputStream), "UTF-8");
+                            appFrameworkService.addRequireExpressionScript(functionName.substring(0, functionName.indexOf(".")), value);
+                        } catch (IOException e) {
+                            log.error("Failed to read appframework require expression: " + functionName + " from " + scriptsDir.getAbsolutePath(), e);
```
mseaton, line 36: Don't catch and log this, just throw a runtime exception.

mseaton (commented):
Generally looks good, one comment

api/src/main/java/org/openmrs/module/pihcore/setup/AppFrameworkExpressionFunctionsSetup.java
cioan: Thanks! All set!

omod/src/main/webapp/resources/scripts/visit/visit.js
```diff
@@ -817,8 +817,8 @@ angular.module("visit", [ "filters", "constants", "encounterTypeConfig", "visitS
                         $scope.encounterDateFormat = sameDate($scope.visit.startDatetime, $scope.visit.stopDatetime) ? "hh:mm a" : "hh:mm a (d-MMM)";
 
                         if ($scope.suppressActions !== true) {
-                          AppFrameworkService.getUserExtensionsFor("patientDashboard.visitActions").then(function (ext) {
-                            $scope.visitActions = ext;
+                          CoreappsService.getUserExtensionsFor("patientDashboard.visitActions", $scope.patientUuid).then(function (ext) {
+                              $scope.visitActions = ext;
```
mseaton, line 821: One thing here - this only takes in the patient, but shouldn't this also take in the visit optionally?  Without the visit, this will always return actions based on the current active visit.  But if we view a "past visit" within the visit note, shouldn't the actions that are available be based on that past visit and not on the currently active visit?
@cioan / @mogoodrich ?

cioan: @mseaton , then we would need to modify the REST controller to accept the visit parameter?
https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/java/org/openmrs/module/coreapps/web/controller/CoreappsRestController.java#L92

mseaton: Presumably, yes, as well as the appcontextmodelgenerator.  Maybe this is a separate ticket, we should discuss.

cioan: @mseaton , I have added the visit parameter to the REST request

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -2163,6 +2164,7 @@ private void enableMCHForms() {
                     and(sessionLocationHasTag("Maternal and Child Location"),
                             visitDoesNotHaveEncounterOfType(PihEmrConfigConstants.ENCOUNTERTYPE_ANC_INTAKE_UUID),
                             visitDoesNotHaveEncounterOfType(PihEmrConfigConstants.ENCOUNTERTYPE_ANC_FOLLOWUP_UUID),
+                            patientDoesNotHaveEncounterOfTypeDuringProgramEnrollment(PihEmrConfigConstants.ENCOUNTERTYPE_ANC_INTAKE_UUID, PihEmrConfigConstants.PROGRAM_PREGNANCY_UUID),
```
mseaton, line 2167: The supporting code looks good, but not sure if this expression will work for all cases?  There is no notion of time here - the implementation below appears to check the patient's active program enrollments.  But what if this is opened in a retrospective context?

Presumably what we are trying to figure out is relative to a particular date - either the current date if no visit is present in the model, or the visit date if it is present in the model.  And as of that date:
* Is the patient enrolled in a particular program on that date
* Did they have an encounter of a particular type at any point between the dateEnrolled and dateCompleted of that enrollment (whether that enrollment is currently active or not).

omod/src/main/webapp/resources/scripts/services/configService.js
```diff
@@ -17,4 +24,23 @@ angular.module('configService', ['ngResource'])
...
-    }]);
\ No newline at end of file
+    }])
+
+    .factory('CoreappsService', [ "ExtensionResource", function(ExtensionResource, AppResource) {
+        return {
+            // returns a promise
+            getUserExtensionsFor: function(extensionPoint, patient, visit) {
+                // TODO handle multiple pages
+                var extensionPointId = extensionPoint.uuid || extensionPoint;
```
mseaton, line 34: There is no `uuid` on an extensionPoint.  It is just a string.

api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
cioan: OK, then we would need to expand the appcontextmodel to include all programs the patient has ever been enrolled. I was working under the impression that this requirement of limiting the number of encounter types applies to only active program enrollments. 
Why did you limit here the list of programs included in the model to only activePrograms?
https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/java/org/openmrs/module/coreapps/contextmodel/AppContextModelGenerator.java#L106

PR #578, opened 2025-05-22
UHM-8650 - Ensure currently selected visit is able to be passed into …

mseaton:
…context to retrieve visit actions


cioan (approved):
Thanks @mseaton !

PR #580, opened 2025-05-29
SL-940: ED triage encounter should not keep the visit open

cioan:


mogoodrich (approved):
Great, thanks @cioan !

cioan:
Thanks @mogoodrich !

PR #584, opened 2025-07-30
SL-983: remove entries older than 12 hours from the MCOE triage queue

cioan:
@mseaton , please let me know if this is the right direction for removing timeout entries from the MCOE triage queue. I will add unit tests and more configuration if necessary? Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
@@ -1,5 +1,6 @@
 package org.openmrs.module.pihcore.task;
 
+import org.apache.commons.lang.time.DateUtils;
 import org.apache.commons.lang.time.StopWatch;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.pihcore.SierraLeoneConfigConstants;
@@ -11,9 +12,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
```diff
@@ -0,0 +1,106 @@
...
+            stopWatch.stop();
+            log.info("Execution completed in {} - {} patients removed", stopWatch, removedPatientsCount);
+        } catch (Exception e) {
+            log.error("Error executing " + getClass(), e);
+        } finally {
+            isExecuting = false;
+        }
+    }
+
+    private Queue getQueue() throws Exception {
```
mseaton, line 57: Isn't this unnecessarily complex?  Why not just retrieve queue by uuid?  The MCOE Triage Queue uuid is a fixed value - 3113f164-68f0-11ee-ab8d-0242ac120002

api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
```diff
@@ -0,0 +1,106 @@
...
+        return queues.get(0);
+    }
+
+    private int removeTimedOutPatients(Queue queue) {
+        int numPatientsRemoved = 0;
+
+        QueueServicesWrapper queueServices = Context.getRegisteredComponents(QueueServicesWrapper.class).get(0);
+        QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+        criteria.setQueues(Collections.singletonList(queue));
+        criteria.setIsEnded(false);
```
mseaton, line 81: Rather than do the date calculations and comparisons in Java below, why not set `startedOnOrBefore` to this criteria instance, and set it to the current date - the configured minutes?  Then you can just end all of the retrieved queue entries.

api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+
+public class PihRemovePatientsFromMCOEQueue implements Runnable {
+
+    private static final Long QUEUE_TIMEOUT_MINUTES = 720L;
```
mseaton, line 21: If the ticket description is to end queues > 12 hours, why don't we configure this in hours rather than in minutes?  I understand that minutes gives more granular flexibility, but hours is more easy to follow and read, if that's what the actual requirement is.  I don't see us changing this in the future to a fraction of an hour...

cioan:
Thanks @mseaton ! I have addressed your comments above. Also added an unit test. Thanks!

PR #590, opened 2025-09-02
HAI-736: Afficher le site auquel appartient un patient lors de la recherche

louidorjp:
Please review this PR

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/service/PihCoreServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/service/PihCoreServiceImpl.java
@@ -25,6 +25,7 @@ import org.openmrs.Encounter;
 import org.openmrs.Obs;
 import org.openmrs.Order;
 import org.openmrs.Patient;
+import org.openmrs.PatientIdentifier;
 import org.openmrs.PatientProgram;
 import org.openmrs.Person;
 import org.openmrs.PersonAttribute;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/service/PihCoreServiceImpl.java
```diff
@@ -208,9 +209,19 @@ public void updateHealthCenter(Patient patient) {
                 // Retrieve all patient programs for this patient that are linked to the HIV program
                 List<PatientProgram> patientPrograms = programWorkflowService.getPatientPrograms(patient, hivProgram, null, null, null, null, false);
                 if (!patientPrograms.isEmpty()) {
+                    // Get the most recent HIV program for the patient by selecting the one with the latest enrollment date
+                    PatientProgram mostRecentProgram = patientPrograms.stream()
+                        .max(Comparator.comparing(PatientProgram::getDateEnrolled))
+                        .orElse(null);
                     // per documentation, the "addAttribute" function will create the attribute if needed, or otherwise override any existing attribute value
-                    patient.addAttribute(new PersonAttribute(healthCenter,patientPrograms.get(0).getLocation().getId().toString()));
+                    patient.addAttribute(new PersonAttribute(healthCenter,mostRecentProgram.getLocation().getId().toString()));
```
mogoodrich, line 217: Looks correct, though it would be good if could confirm with a unit test?

api/src/main/java/org/openmrs/module/pihcore/service/PihCoreServiceImpl.java
```diff
@@ -208,9 +209,19 @@ public void updateHealthCenter(Patient patient) {
...
                     patientService.savePatient(patient);
+                    
+                } else {
+                    for (PatientIdentifier pid : patient.getIdentifiers()) {
```
mogoodrich, line 221: From what @dmdesimone  said in the ticket, I think we want to use the location of the most recent registration encounter as the fallback, not the identifier location?

We can do this by using a method from the Encounter Service to fetch encounters. I believe this method should work:

https://github.com/PIH/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/EncounterService.java#L182

We would create an "EncounterSearchCriteria" object and set the patient and encounter type on it.  The tests provide some examples of how to use it:

https://github.com/PIH/openmrs-core/blob/master/api/src/test/java/org/openmrs/api/EncounterServiceTest.java#L1086

You should be able to look up the encounter type using this method:

https://github.com/PIH/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/EncounterService.java#L268

... getting the "PATIENT_REGISTRATION" encounter type uuid from the PihEmrConfigConstants

Then we'd sort by encounter date similar to how we sort program by enrollment date, and take the most recent.

mogoodrich (commented):
Thanks @louidorjp !  Generally looks good, I added a few comments.  It would also be helpful to add some tests of this new logic.  I'm happy to help with that.

PR #597, opened 2025-09-26
SL-983: one duplicate entry should not prevent removing other entries from the queue

cioan:
@mseaton , this change would allow removing from the queue valid entries which are not duplicate. Maybe because of the previous bug addressed in https://openmrs.atlassian.net/browse/O3-4985 duplicate entries may be present in the queue and therefore the entire code below was not executing if the duplicate entry happened to be the first entry processed.

### DIFF
--- a/api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
+++ b/api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
@@ -91,12 +91,11 @@ public class PihRemovePatientsFromMCOEQueue implements Runnable {
             entry.setEndedAt(currentTime);
             try {
                 queueServices.getQueueEntryService().saveQueueEntry(entry);
+                log.warn("Removing patient {} from queue after {} minutes", entry.getPatient().getUuid(), activeMinutes);
+                numPatientsRemoved++;
             } catch (Exception e) {
                 log.error("Failed to remove patient: " + entry.getPatient().getUuid() + " from queue.", e);
-                continue;
             }
-            log.warn("Removing patient {} from queue after {} minutes", entry.getPatient().getUuid(), activeMinutes);
-            numPatientsRemoved++;
         }
 
         return numPatientsRemoved;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
```diff
@@ -88,11 +88,15 @@ private int removeTimedOutPatients(Queue queue) {
...
             entry.setEndedAt(currentTime);
-            queueServices.getQueueEntryService().saveQueueEntry(entry);
+            try {
+                queueServices.getQueueEntryService().saveQueueEntry(entry);
+            } catch (Exception e) {
+                log.error("Failed to remove patient: " + entry.getPatient().getUuid() + " from queue.", e);
+                continue;
+            }
+            log.warn("Removing patient {} from queue after {} minutes", entry.getPatient().getUuid(), activeMinutes);
             numPatientsRemoved++;
```
mseaton, line 99: Personally I don't like using the `continue;` construct, as it is confusing and hard to nest.  Can we instead just move these 2 statements into the try block, and then if an exception is caught we simply log that?

api/src/main/java/org/openmrs/module/pihcore/task/PihRemovePatientsFromMCOEQueue.java
```diff
@@ -88,11 +88,14 @@ private int removeTimedOutPatients(Queue queue) {
...
             Date startedAt = entry.getStartedAt();
             long activeMinutes = Duration.between(startedAt.toInstant(), currentTime.toInstant()).toMinutes();
-            log.warn("Removing patient {} from queue after {} minutes", entry.getPatient().getUuid(), activeMinutes);
             entry.setEndedAt(currentTime);
-            queueServices.getQueueEntryService().saveQueueEntry(entry);
-            numPatientsRemoved++;
-
+            try {
+                queueServices.getQueueEntryService().saveQueueEntry(entry);
+                log.warn("Removing patient {} from queue after {} minutes", entry.getPatient().getUuid(), activeMinutes);
```
mseaton, line 94: Not sure why you moved this log statement here?

mseaton: I mean from the original code @cioan .  the log statement says "removing", not "removed", which implies the logging is coming prior to the operation.  I am not sure why you moved this to happen _after_ the save.  We can move it there I guess, but then I would expect the message to say "Removed patient ...".

PR #651, opened 2026-05-01
SL-732: add infectious diseases widget

cioan:
@mseaton , this just adds a new widget similar to the PregnancyIndicators widget to the patient dashboard:

 
Screenshot 2026-05-01 at 2 56 44 PM

### DIFF
--- a/omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
+++ b/omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
@@ -1,0 +1,22 @@
+<style>
+    .red {
+        color: red;
+    }
+</style>
+<div class="info-section">
+    <div class="info-header">
+        <i class="${app.icon}"></i>
+        <h3>${ ui.message(app.label) }</h3>
+    </div>
+    <div class="info-body">
+        <% fields?.each { concept, obj -> %>
+            <div style="font-size: smaller;">
+                <span style="font-family: 'OpenSansBold';">${ ui.message(obj.label) }:</span>
+                <span class="${obj.css ? obj.css : ''}">${ obj.obs ?: ui.message("coreapps.none") }</span>
+                <% if (obj.date) { %>
+                    <span style="color: gray;">(${ ui.format(obj.date) })</span>
+                <% } %>
+            </div>
+        <% } %>
+    </div>
+</div>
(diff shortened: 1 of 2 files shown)


mseaton (approved):
LGTM @cioan .

omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
```diff
@@ -0,0 +1,22 @@
...
+        <i class="${app.icon}"></i>
+        <h3>${ ui.message(app.label) }</h3>
+    </div>
+    <div class="info-body">
+        <% fields?.each { concept, obj -> %>
+            <div style="font-size: smaller;">
+                <span style="font-family: 'OpenSansBold';">${ ui.message(obj.label) }:</span>
+                <span class="${obj.css ? obj.css : ''}">${ obj.obs ?: ui.message("coreapps.none") }</span>
+                <% if (obj.date) { %>
+                    <span style="color: gray;">(${ ui.format(obj.date) })</span>
```
mseaton, line 17: I actually like the way this styling looks, but curious about why it was applied, since it isn't shown like that in the requirements and is inconsistent with the way the other widgets are displayed.

I also wouldn't use `font-family: 'OpenSansBold'` to make something bold - it seems like a lot of this PR is AI generated - is this what AI is recommending?

omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
```diff
@@ -0,0 +1,22 @@
...
+    }
+</style>
+<div class="info-section">
+    <div class="info-header">
+        <i class="${app.icon}"></i>
+        <h3>${ ui.message(app.label) }</h3>
+    </div>
+    <div class="info-body">
+        <% fields?.each { concept, obj -> %>
+            <div style="font-size: smaller;">
```
mseaton, line 13: I do like the conciseness of this - having 3 rows of data instead of 6, but that's not what the ticket says.  Was this agreed to change it like this?

omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
cioan: @mseaton , the font-family: 'OpenSansBold'  is the same as used in pregnancyIndicator widget: 

https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/fragments/dashboardwidgets/pregnancyIndicators.gsp#L17

Should I just use the <b> tag instead of the above? Other suggestions?

Yes, I think too that not only looks better having the date of the test next to the test result, but also saves us space in an increasingly crowded dashboard. I will see what Fiona thinks about this change. It's easy enough to add another line if needed.

omod/src/main/webapp/fragments/dashboardwidgets/infectiousDiseases.gsp
cioan: Please see my comment above. I did not get a chance to review this suggestion with Fiona. It would be easy enough to add the test date on a separate line if needed.

PR #652, opened 2026-05-05
SL-1324: add born during visit attribute

cioan:


api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
```diff
@@ -241,6 +244,13 @@ Visit createVisit(Patient baby, Date startDatetime, Encounter motherEncounter) {
             visit.setLocation(adtService.getLocationThatSupportsVisits(motherEncounter.getLocation()));
             visit.setStartDatetime(startDatetime);
             visit.setVisitType(visitService.getVisitTypeByUuid(PihEmrConfigConstants.VISITTYPE_CLINIC_OR_HOSPITAL_VISIT_UUID));
+            VisitAttributeType visitAttributeType = visitService.getVisitAttributeTypeByUuid(BORN_DURING_VISIT_ATTRIBUTE_TYPE);
+            if (visitAttributeType != null) {
+                VisitAttribute visitAttribute = new VisitAttribute();
+                visitAttribute.setAttributeType(visitAttributeType);
+                visitAttribute.setValueReferenceInternal("true");
```
mogoodrich: Not sure if there's a way to set this without using the "valueReferenceInternal" method?

mogoodrich (approved):
Generally LGTM, though I wasn't sure if the API provides a way to directly set the value without having to use the "setValueReferenceInternal" method?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/RegisterBabyAction.java
cioan: No problem, @mogoodrich . `visitAttribute.setValue(new Boolean(true));` works as well.

PR #657, opened 2026-05-22
UHM-9313: Unable to add past lab results outside of active visit

mogoodrich:


api/src/main/java/org/openmrs/module/pihcore/apploader/CustomAppLoaderFactory.java
```diff
@@ -1010,7 +1011,7 @@ private void enableDeathCertificate() {
                 "mirebalais.deathCertificate.death_certificate",
                 "fas fa-fw fa-times-circle",
                 "link",
-                enterSimpleHtmlFormLink(PihCoreUtil.getFormResource("deathCertificate.xml")),
+                enterSimpleHtmlFormLinkWithoutVisit(PihCoreUtil.getFormResource("deathCertificate.xml")),
```
mogoodrich, line 1014: I was looking for other forms that would potentially have the same problem and this was the only one I found... I did not test, but since this isn't used seems very low risk.

cioan (approved):
LGTM!

mseaton (approved):

PR #658, opened 2026-05-22
UHM-9208: Restrict changing Visit Location

mogoodrich:


mseaton (approved):
OK, we should keep track of all of these places in the event that this moves into a new place in core in the next few months.

mogoodrich:
Definitely true @mseaton.  I think/hope this is the only other one we need to implement for now.

PR #663, opened 2026-06-02
UHM-9280: Optimize queries for concepts in LabResultsFragmentController

chibongho:
Reimplement the `getLabCategoriesList()` function so that when we traverse through the concept hierarchy of he `labCategoriesSet`, we batch the queries for concepts by uuids at each depth level. 

I had Claude code reviewed this and it mentioned that the old code includes retired concepts. That was probably unintended, so I didn't do that here. If that's not the case I can change the `ConceptSearchCriteriaBuilder` to call `includeRetired(true)`

### DIFF
--- a/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/LabResultsFragmentController.java
+++ b/omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/LabResultsFragmentController.java
@@ -150,7 +150,7 @@ public class LabResultsFragmentController {
 
         while (!frontier.isEmpty()) {
             List<Concept> concepts = conceptService.getConcepts(
-                    new ConceptSearchCriteriaBuilder().addUuids(frontier.toArray(new String[0])).build());
+                    new ConceptSearchCriteriaBuilder().addUuids(frontier.toArray(new String[0])).includeRetired(true).build());
 
             visited.addAll(frontier);
             frontier.clear();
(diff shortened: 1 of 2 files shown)


mogoodrich (changes requested):
This doesn't seem to do a multiple-level search anymore, which I believe we need to do?

And, yes we need to include retired concepts--even if the concept is retired going forward, if there is an old result under that concept we want to display it.

mseaton (commented):
I am going to need you to explain this code to me @chibongho .  Easier to do a review together than have a back-and-forth.

chibongho:
Ok, updated to include retired concepts.

I did forgot to update the tests to reflect my changes. I just pushed an update to fix them and have a test for multi-level search.

omod/src/main/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/LabResultsFragmentController.java
```diff
@@ -138,36 +139,28 @@ else if (patientConfig instanceof String) {
...
-            // Check if already visited (prevents cycles and redundant processing)
-            if (visited.contains(labCategoriesSet)) {
-                return labCategories;
-            }
+        Set<String> visited = new HashSet<>();
+        Set<String> frontier = new HashSet<>();
+        frontier.add(labCategoriesSet);
 
-            visited.add(labCategoriesSet);
+        while (!frontier.isEmpty()) {
```
mogoodrich, line 151: Oh, I see, this is the iterative block?

chibongho:
Closing this after discussion. Implement a function to recursively get concept set members instead in pihapps: https://github.com/PIH/openmrs-module-pihapps/pull/33/changes

PR #665, opened 2026-06-03
UHM-9280: Optimize queries for concepts in LabResultsFragmentController

chibongho:


omod/src/test/java/org/openmrs/module/pihcore/fragment/controller/dashboardwidgets/LabResultsFragmentControllerTest.java
chibongho, line 1: Deleting this file as the tests are now covered in https://github.com/PIH/openmrs-module-pihapps/pull/33/changes

mseaton (approved):
Not sure why we don't just add the Set of concepts to the model rather than the set of uuids, but fine I guess

PR #668, opened 2026-06-10
SL-1290 Add post submission action to schedule appointment

chibongho:


mogoodrich (commented):
Could we add a couple tests around this?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+public class ScheduleAppointmentAction implements CustomFormSubmissionAction {
+
+    static final String SCHEDULE_APPOINTMENT_CONCEPT = "PIH:Schedule appointment";
+    static final String APPOINTMENT_GROUP_CONCEPT    = "PIH:Scheduled visit construct";
+    static final String APPOINTMENT_DATETIME_CONCEPT = "PIH:Return visit datetime";
+    static final String APPOINTMENT_LOCATION_CONCEPT = "PIH:Location";
+    static final String APPOINTMENT_SERVICE_CONCEPT  = "PIH:Appointment service";
+    static final String APPOINTMENT_TYPE_CONCEPT     = "PIH:Visit type";
+    static final String APPOINTMENT_PROVIDER_CONCEPT = "PIH:Provider name non-coded";
+    static final String APPOINTMENT_NOTES_CONCEPT    = "PIH:GENERAL FREE TEXT";
```
mseaton, line 40: Presumably these are concepts @lnball recommended?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void applyAction(FormEntrySession session) {
+        if (session.getContext().getMode() != Mode.ENTER) {
+            return;
```
mseaton, line 45: Something worth confirming.  Presumably we could check whether the "old" values match an existing appointment exactly for the patient, and if so, and if the new values are different, then we'd want to update the appointment accordingly.  I could see not wanting to do so if it has diverged, but if it hasn't then I'd think we'd want to update it.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        if (!isScheduleAppointmentYes(session)) {
+            return;
```
mseaton, line 51: And void it in this case if it was previously scheduled via HFE, unchanged, and then unscheduled.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        if (appointmentGroup == null) {
+            return;
+        }
+
+        Date startDateTime = getDatetimeValue(appointmentGroup, APPOINTMENT_DATETIME_CONCEPT);
+        String locationText = getTextValue(appointmentGroup, APPOINTMENT_LOCATION_CONCEPT);
+        String serviceUuid  = getTextValue(appointmentGroup, APPOINTMENT_SERVICE_CONCEPT);
+        Concept typeValue   = getCodedValue(appointmentGroup, APPOINTMENT_TYPE_CONCEPT);
+        String providerText = getTextValue(appointmentGroup, APPOINTMENT_PROVIDER_CONCEPT);
+        String notes        = getTextValue(appointmentGroup, APPOINTMENT_NOTES_CONCEPT);
```
mseaton: Can we be consistent in our variable naming?  "Text" vs "Value" vs "Uuid"?  Can we just call each "Ref"?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+
+        if (startDateTime == null || locationText == null || serviceUuid == null) {
+            return;
```
mseaton: So we just silently skip?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        Location location = HtmlFormEntryUtil.getLocation(locationText);
+        AppointmentServiceDefinition service = getAppointmentService(serviceUuid);
+
+        if (location == null || service == null) {
+            return;
```
mseaton: Again, just silently skip?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        Appointment appointment = new Appointment();
+        appointment.setPatient(session.getEncounter().getPatient());
+        appointment.setLocation(location);
+        appointment.setService(service);
+        appointment.setStartDateTime(startDateTime);
+        if (service.getDurationMins() != null) {
+            appointment.setEndDateTime(new Date(startDateTime.getTime() + (long) service.getDurationMins() * 60 * 1000));
```
mseaton, line 86: Why should this be necessary?  It's fine if it _is_ necessary, but shouldn't the appointments api take care of this if this is entirely using data it already has?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        AppointmentKind kind = getAppointmentKind(typeValue);
+        if (kind != null) {
+            appointment.setAppointmentKind(kind);
```
mseaton: Wouldn't you set this on the appointment whether it is null or not?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        if (notes != null) {
+            appointment.setComments(notes);
```
mseaton: Again, wouldn't you set this regardless of whether it is null?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        if (providerText != null) {
+            Provider provider = HtmlFormEntryUtil.getProvider(providerText);
+            if (provider != null) {
```
mseaton: What if providerText is not null and provider is?   Shoudn't providerText be an isBlank check vs a null check?  And shouldn't we fail submission of something is amiss?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        Context.getService(AppointmentsService.class).validateAndSave(appointment);
+    }
+
+    private boolean isScheduleAppointmentYes(FormEntrySession session) {
+        Concept yesConcept = HtmlFormEntryUtil.getConcept("PIH:YES");
+        for (Obs obs : session.getEncounter().getObsAtTopLevel(false)) {
+            if (obs.getConcept().equals(HtmlFormEntryUtil.getConcept(SCHEDULE_APPOINTMENT_CONCEPT))) {
+                return yesConcept.equals(obs.getValueCoded());
```
mseaton, line 110: So this obs is _outside_ of the appointment construct?  And the appointment construct is only created if this is Yes?  Is that normal?

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
```diff
@@ -0,0 +1,171 @@
...
+            return AppointmentKind.WalkIn;
+        }
+        if (concept.equals(HtmlFormEntryUtil.getConcept("CIEL:166997"))) {
+            return AppointmentKind.Virtual;
+        }
+        return null;
+    }
+
+    private AppointmentServiceDefinition getAppointmentService(String uuid) {
+        return Context.getService(AppointmentServiceDefinitionService.class)
```
mseaton: Seriously?  There is no method in the appointments module to get a service definition by uuid??

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
lnball: @chibongho and I worked together on these concepts.  Chi Bong did some concept updates on OCL (a win!)

FYI @mseaton

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
chibongho: I think @fiona855 wanted it to force the user to go to the appointments app to edit the appointment. I made the appointment fields not editable in EDIT mode (via JS).

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
chibongho: You're right, there is. Changing it to call `.getAppointmentServiceByUuid()`.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
chibongho: Yes, this just records whether we want to create an appointment. Not part of the appointment construct. The obs for constructing appointment is only saved if user answers yes.

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
chibongho: The API accepts `startDatetime` and `endDatetime`, but the frontend, following the same format as the O3 form, asks for startDatetime and duration (minutes) instead

mogoodrich (approved):

api/src/main/java/org/openmrs/module/pihcore/htmlformentry/action/ScheduleAppointmentAction.java
mseaton: Still, it is worth discussing.  Doesn't need to hold this up for testing and evaluation, but perhaps bring it up for discussion together.

mseaton (approved):
Looks ok pending my questions/concerns but we can tackle those as enhancements or discussions once this is in place and we have a chance to see how it works.

PR #669, opened 2026-06-22
SL-1322: automatically print after the form loads in the view mode

cioan:
@mseaton , I was able to automatically send the form to the printer after it gets loaded in the view mode:

Screenshot 2026-06-22 at 3 17 00 PM

Please let me know if this looks like the right direction before I go and implement the rest of the formatting requested by the SL team.


mseaton (approved):
LGTM @cioan

cioan:
Thanks @mseaton !

PR #672, opened 2026-07-14
SL-1368: if requested automatically print a form upon submission

cioan:
@mseaton , @mogoodrich , in order to implement this form Save&Print action, I have used the same mechanism we are using to automatically expand the newly saved encounter form. I have just appended **printForm=true** to the returnUrl, and then in the visit.js just parse this parameter and automatically print the form using the print template of the form. 
Once I implemented this, I have noticed a few javascript and Angular errors in the console, and that was caused by the fact that the browsers by default block pop-up windows. So I changed the print function to load the printing template in an iFrame, which works much better than the previous solution with the dialog box. Both printing methods, from the visit dashboard or by clicking Print&Save on the form works fine with no errors now.
Here are the changes I made to the maternalDischarge form:
https://github.com/PIH/openmrs-config-pihsl/pull/326


mseaton (approved):
Fine with me if it works.

mogoodrich (approved):
Thanks @cioan !

PR #676, opened 2026-07-22
UHM-9394: Fix PIH Core Build after removal of web layer from api build

mogoodrich:


mseaton (commented):
I guess this is sort of fine, but I worry that this will bite us in the future and we will start getting weird errors if we forget to update this.

If we can instead just move the biometric stuff into the omod layer, that would seem preferable.

PR #683, opened 2026-08-13
TASKS-585 - Simplify pih config loading and support empty "site" property

mseaton:
This PR makes a few changes to config loading:

1. Adds new convenience methods "isCountry" and "isSite" to Config that are null-safe, and changes code to use these new methods rather than previous raw .equals comparisions.

2. Adds a clear error to the log file if the server fails to start up due to invalid or missing pih config that looks like this:

ERROR - Config.<init>(35) |2026-08-13T10:34:15,201| 
***************************************************************************
* PIH CORE MODULE FAILED TO START
* IllegalStateException: pih config is not defined
***************************************************************************
java.lang.IllegalStateException: pih config is not defined
	at org.openmrs.module.pihcore.config.ConfigLoader.load(ConfigLoader.java:45) ~[pihcore-api-2.2.0-SNAPSHOT.jar:?]

Previously, this was a more opaque and hard to see error that one had to deduce by digging through the log and looking for a failure to parse pih config json during startup.

3. Simplifies the config loading mechanisms by:
* Removing the pih.config.dir configuration parameter, which we don't use afaik (it's always configuration/pih in app data dir
* Removing all concept of a "default" configuration that is set overall, or in a site configuration in favor of an explicitly configured value always

4. Updates to a handful of places (id generator and metadata mapping setup) to not fail if no site is found in the config, but rather just skip over the configuration, in order to allow the system to successfully start up without a matching "site".

@mogoodrich there will be corresponding commits to leverage this in our emr projects.


mogoodrich (approved):
LGTM, thanks @mseaton

